PR #7, opened 2022-08-08
Task/add cars categories component

elador13:
See - https://app.asana.com/0/1202525679663009/1202720798232787/f


frontend/src/types/brand.type.ts
```diff
@@ -0,0 +1,8 @@
+export type Brand = {
+  name: string;
+  marka_id: number;
+  eng: string;
+  country_id: number;
+  category_id: number;
+  logo: string;
+};
```
oleksandr-danylchenko: The `Brand` type should correspond to the one in the DB. It should be more than enough to have only the `id`, `logo_url`, and `autoria_code` in this type.

https://dbdiagram.io/d/62e95fbbf31da965e865f0c9
image

frontend/src/components/cars-categories/cars-categories.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const CarsCategories: FC = () => {
+  const [brands, setBrands] = useState<Brand[]>([]);
+  const [activeId, setActiveId] = useState<number>();
+
+  useEffect(() => {
+    if (mockBrands.length) {
+      setBrands(mockBrands);
+    }
+  }, []);
```
oleksandr-danylchenko: You can simply inject the `mockBrands` as the initial state for the `brands`. Because for fetching we won't use the `useEffect` + `useState`, but RTKQ. So there's no need to complicate the component for now

frontend/src/components/cars-categories/styles.module.scss
```diff
@@ -0,0 +1,31 @@
+.container {
+  position: absolute;
```
oleksandr-danylchenko: Why does the container need to be `absolute`? This brands list should be part of the homepage and included in its natural elements flow. So I think it can use the default `position:static`

frontend/src/components/cars-categories/styles.module.scss
```diff
@@ -0,0 +1,31 @@
...
+  display: flex;
+  flex-direction: row;
```
oleksandr-danylchenko: `flex-direction: row` is the default value. Therefore, there's no need to explicitly specify it in the code

frontend/src/components/cars-categories/cars-categories.tsx
```diff
@@ -0,0 +1,43 @@
...
+        >
+          <Link
+            to="#"
+            onClick={(): void => setActiveId(brand.marka_id)}
+            className={styles.navLink}
+          >
+            <img className={styles.logo} src={brand.logo} alt={brand.name} />
+          </Link>
+        </div>
+      ))}
```
oleksandr-danylchenko: I would expect the whole tile to the clickable, not only the image. Therefore, in my view, we can use only the `Link` component, w/o the `div` wrapper, and style it as the brand tile

frontend/src/components/cars-categories/cars-categories.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <div className={styles.container}>
+      {brands.map((brand) => (
+        <div
+          key={brand.marka_id}
+          className={
+            brand.marka_id === activeId
+              ? styles.linkWrapper + ' ' + styles.active
+              : styles.linkWrapper
```
oleksandr-danylchenko: I don't think that we need to store the active link id state. Because after the click, we would be redirected to another page, and that state change won't be even seen.

When I said "Desing the active state", I meant the [CSS's `:active`](https://developer.mozilla.org/en-US/docs/Web/CSS/:active) :slightly_smiling_face: 
So you just need to have not the `styles.active`, but `.linkWrapper:active` (or any other name of the class for the element which will show the brand tile)

oleksandr-danylchenko (changes requested):
I see where it's going, good job! But there're some simplifications can be done as well :raised_hands:

oleksandr-danylchenko:
Also, we need to name all the files in `kebab-case`. So the `mockBrands.ts` should be the `mock-brands.ts`

frontend/src/components/cars-categories/cars-categories.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { FC, useState } from 'react';
+import { Link } from 'react-router-dom';
+
+import { mockBrands } from '@components/cars-categories/mock-brands';
+
+import { Brand } from '../../types/brand.type';
+import styles from './styles.module.scss';
+
+const CarsCategories: FC = () => {
+  const [brands] = useState<Brand[]>([...mockBrands]);
```
oleksandr-danylchenko: `[...mockBrands]` буде повністю рівним `mockBrands` :slightly_smiling_face: 
Тому можна просто передати `mockBrands` в `useState`

oleksandr-danylchenko (approved):
Now the component looks as expected, thanks!

PR #10, opened 2022-08-09
Task/create users seeds

stanislav-ostr:
TASK
Create users factory with seeds 

users data, passwords:
https://docs.google.com/spreadsheets/d/1utbBxZIhjhmib6YRhmVlxx6NmbspofqGBjm-igGRiFk/edit?usp=sharing


backend/prisma/schema.prisma
```diff
@@ -8,11 +8,40 @@ datasource db {
...
-  id       String  @id @default(uuid())
-  email    String  @unique @db.VarChar(80)
-  password String  @db.VarChar(300) // Length of the hashed password
+  id            String         @id @default(uuid())
   /// Full name of a user
-  name     String  @db.VarChar(200)
-  phone    String? @db.VarChar(20)
-  location String?
+  name          String         @db.VarChar(200)
+  sex           Sex
```
oleksandr-danylchenko: In the modern world making the `sex` field mandatory can lead to the Twitter scandal :sweat_smile:. So I would like to make it optional too. To make it optional, you can create a new migration. In Prisma philosophy, they don't delete or revert migrations, but just keep adding new ones

oleksandr-danylchenko (dismissed):
All the changes look sensible to me :+1: 
I would like to change one single DB field to be optional and then the approval will be 100% justified!

backend/prisma/schema.prisma
```diff
@@ -8,11 +8,40 @@ datasource db {
 }
 
 model User {
```
RomanMahola, line 10: It also looks good for me 
Only one thing to mention. Next time you can use faker with factory pattern. It would satisfy us if we will need more then 1000 users for example 

Refs:
https://fakerjs.dev/
https://refactoring.guru/design-patterns/factory-method

RomanMahola (dismissed):

oleksandr-danylchenko (approved):
LGTM!

PR #13, opened 2022-08-10
fix errors after moved password to user_security

stanislav-ostr:


oleksandr-danylchenko (approved):
Looks as I would expect it to work :+1:

PR #14, opened 2022-08-10
Task/add slider to cars categories section

elador13:
See - https://app.asana.com/0/1202525679663009/1202720798232791/f


frontend/src/components/cars-categories/cars-categories.tsx
```diff
@@ -2,21 +2,40 @@ import React, { FC, useState } from 'react';
...
 
 import { mockBrands } from '@components/cars-categories/mock-brands';
+import { Mousewheel, Navigation, SwiperOptions } from 'swiper';
+import { Swiper, SwiperSlide } from 'swiper/react';
 
 import { Brand } from '../../types/brand.type';
+import 'swiper/css';
+import 'swiper/css/navigation';
+import 'swiper/css/mousewheel';
 import styles from './styles.module.scss';
```
oleksandr-danylchenko, line 12: It's recommended to put the 3rd party imports before the project-related imports. So I would suggest putting these styles right after the `import { Swiper, SwiperSlide } from 'swiper/react';`

```suggestion
import { Swiper, SwiperSlide } from 'swiper/react';
import 'swiper/css';
import 'swiper/css/navigation';
import 'swiper/css/mousewheel';

import { Brand } from '../../types/brand.type';
import styles from './styles.module.scss';
```

frontend/src/components/cars-categories/cars-categories.tsx
```diff
@@ -2,21 +2,40 @@ import React, { FC, useState } from 'react';
 import { Link } from 'react-router-dom';
 
 import { mockBrands } from '@components/cars-categories/mock-brands';
+import { Mousewheel, Navigation, SwiperOptions } from 'swiper';
+import { Swiper, SwiperSlide } from 'swiper/react';
 
 import { Brand } from '../../types/brand.type';
+import 'swiper/css';
+import 'swiper/css/navigation';
+import 'swiper/css/mousewheel';
```
oleksandr-danylchenko: It's recommended to put the 3rd party imports before the project-related imports. So I would suggest putting these styles right after the `import { Swiper, SwiperSlide } from 'swiper/react';`

```suggestion
import { Swiper, SwiperSlide } from 'swiper/react';
import 'swiper/css';
import 'swiper/css/navigation';
import 'swiper/css/mousewheel';

import { Brand } from '../../types/brand.type';
import styles from './styles.module.scss';
```

frontend/src/components/cars-categories/cars-categories.tsx
```diff
@@ -2,21 +2,40 @@ import React, { FC, useState } from 'react';
...
+    slidesPerView: 8,
+    loop: true,
+    mousewheel: true,
+    navigation: true,
+    modules: [Navigation, Mousewheel],
+  };
+
   return (
-    <div className={styles.container}>
+    <Swiper className={styles.container} {...swiperParams}>
```
oleksandr-danylchenko: The static object like this one should be declared either **_outside_** of the component or **_in the `useMemo`_**. It's because on each render this object will be recreated and will inevitably lead to the `Swiper`'s rerender

oleksandr-danylchenko (changes requested):
Looks good! Just need a slight refactoring and it'll be ready to be shipped

oleksandr-danylchenko (approved):
LGTM. I think that the large diff is caused by the merging of the different branches inside of this one. But IDE says that the actual diff is only in these PRs, so we should be quite safe
image

PR #16, opened 2022-08-10
Add reset password functionality backend

ashonuyak:
See - https://app.asana.com/0/1202525679663009/1202720798232783/f
Mailtrap section in env.enum is named as "Mailtrap" because it contains credentials to Mailtrap mail access.
We use nodemailer library to send mails to Mailtrap website.


backend/src/controllers/auth/auth.controller.ts
```diff
@@ -21,4 +21,70 @@ const signupLocal = async (
   }
 };
 
-export { signupLocal };
+const resetPasswordRequest = async (
+  req: Request,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
```
oleksandr-danylchenko: We shouldn't use the loosely typed `Request` type. Please refer to the more type-safe alternatives in the `types/controller`. Or, if there's no type-safe `Request` alternative that supports the `params` - add it please :pray:

backend/src/routes/auth/auth.router.ts
```diff
@@ -7,4 +7,14 @@ const authRouter = Router();
...
 
+authRouter.get(
+  `${PATH}/local/reset-password/:email`,
+  authController.resetPasswordRequest,
+);
+authRouter.get(
+  `${PATH}/local/reset-password`,
+  authController.resetPasswordCheckToken,
+);
+authRouter.post(`${PATH}/local/reset-password`, authController.resetPassword);
```
oleksandr-danylchenko: From my experience, I would recommend having different paths for all the 3 endpoints. It'll make our like much easier during debugging to clearly understand what endpoint is responsible for what

backend/src/routes/auth/auth.router.ts
```diff
@@ -7,4 +7,14 @@ const authRouter = Router();
 
 authRouter.post(`${PATH}/local/signup`, authController.signupLocal);
 
+authRouter.get(
+  `${PATH}/local/reset-password/:email`,
+  authController.resetPasswordRequest,
+);
```
oleksandr-danylchenko: This controller modifies the DB, so it must be the `POST` request
image

backend/src/controllers/auth/auth.controller.ts
```diff
@@ -21,4 +21,70 @@ const signupLocal = async (
...
+  try {
+    const requestPasswordResetService = await authService.requestPasswordReset(
+      req.params.email,
+    );
+    res.json(requestPasswordResetService);
```
oleksandr-danylchenko: `requestPasswordReset` definitely returns not the `requestPasswordResetService` :slightly_smiling_face:

backend/src/helpers/mailtrap/mailtrap.ts
```diff
@@ -0,0 +1,41 @@
...
+  template: string,
+): Promise<void> => {
+  const transporter = nodemailer.createTransport({
+    host: ENV.MAILTRAP.EMAIL_HOST,
+    port: 587,
+    auth: {
+      user: ENV.MAILTRAP.EMAIL_USERNAME,
+      pass: ENV.MAILTRAP.EMAIL_PASSWORD,
+    },
+  });
```
oleksandr-danylchenko, line 19: Can we share the `transport` between all the `sendEmail` invocations? Or should it be created for each email?

backend/src/helpers/mailtrap/template/requestResetPassword.handlebars
```diff
@@ -0,0 +1,30 @@
...
+              <p style="text-align:left">
+              Sincerely,<br>Autoline Team
+              </p>
+            </td>
+          </tr>
+        </tbody>
+      </table>
+    </center>
+  </body>
+</html>
```
oleksandr-danylchenko: Thanks for thinking about the styles in front

backend/src/services/auth/auth.service.ts
```diff
@@ -31,4 +34,102 @@ const signupLocal = async (user: UserCreateInput): Promise<AuthResponseDto> => {
...
+  });
+  if (!user) throw new Error('User with such email does not exist');
+
+  const resetToken = jwt.sign(
+    {
+      id: user.id,
+    },
+    ENV.JWT.SECRET!,
+    { expiresIn: '1h' },
+  );
```
oleksandr-danylchenko: JWT has the convention that the `user.id` is stored under the `sub` property

backend/src/services/auth/auth.service.ts
```diff
@@ -31,4 +34,102 @@ const signupLocal = async (user: UserCreateInput): Promise<AuthResponseDto> => {
...
+
+  await prisma.user_Security.update({
+    where: { user_id: user.id },
+    data: { password_change_token: resetToken },
+  });
+
+  const link = `localhost:${ENV.APP.SERVER_PORT}${ENV.API.V1_PREFIX}/auth/local/reset-password?token=${resetToken}`;
+
```
oleksandr-danylchenko: We **mustn't** refer to the `localhost` from the source code. We should use the `ENV.APP.SERVER_HOST`. And the env variable in `backend.env` can be set as `localhost`

backend/src/services/auth/auth.service.ts
```diff
@@ -31,4 +34,102 @@ const signupLocal = async (user: UserCreateInput): Promise<AuthResponseDto> => {
...
+    });
+    throw new Error('Invalid password reset token');
+  }
+
+  await prisma.user_Security.update({
+    where: { user_id: payload.id },
+    data: { password_change_token: null },
+  });
+
+  return payload.id;
```
oleksandr-danylchenko: The duplication of the token deletion can be avoided like this: 

```suggestion
  // Resets password change token either if activation was successful or failed
  await prisma.user_Security.update({
    where: { user_id: payload.id },
    data: { password_change_token: null },
  });
  
  if (token !== userSecurity.password_change_token) {
    throw new Error('Invalid password reset token');
  }

  return payload.id;
```

backend/src/services/auth/auth.service.ts
```diff
@@ -31,4 +34,102 @@ const signupLocal = async (user: UserCreateInput): Promise<AuthResponseDto> => {
...
+  const userSecurity = await prisma.user_Security.findUnique({
+    where: { user_id: id },
+  });
+  if (!userSecurity) {
+    throw new Error('User does not exist');
+  }
+  const isSamePassword = await bcrypt.compare(password, userSecurity.password);
+  if (isSamePassword) {
+    throw new Error('Same passwords error');
+  }
```
oleksandr-danylchenko: This code lacks some breathing space :slightly_smiling_face: 

```suggestion
  const user = await prisma.user.findUnique({ where: { id } });
  if (!user) {
    throw new Error('User does not exist');
  }
  
  const userSecurity = await prisma.user_Security.findUnique({
    where: { user_id: id },
  });
  if (!userSecurity) {
    throw new Error('User does not exist');
  }
  
  const isSamePassword = await bcrypt.compare(password, userSecurity.password);
  if (isSamePassword) {
    throw new Error('Same passwords error');
  }
```

backend/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../tsconfig.json",
-  "include": ["src"],
+  "include": ["src", "src/helpers/mailtrap/mailtrap.ts"],
```
oleksandr-danylchenko: Why do we need to explicitly include smth into types as it's already in the `src` folder?

backend/src/services/auth/auth.service.ts
```diff
@@ -31,4 +34,102 @@ const signupLocal = async (user: UserCreateInput): Promise<AuthResponseDto> => {
...
+    data: { password: hash },
+  });
+
+  sendEmail(
+    user.email,
+    'Password Reset Successfully',
+    {
+      name: user.name,
+    },
+    './template/resetPassword.handlebars',
```
oleksandr-danylchenko: I would name the template as `resetPasswordConfirmation.handlebars` to be more explicit

oleksandr-danylchenko (changes requested):
I see the direction and I like it! But this PR needs some additional polishing to be approved :slightly_smiling_face:

backend/src/common/enums/app/env.enum.ts
```diff
@@ -28,6 +32,12 @@ const ENV = {
   API: {
     V1_PREFIX: '/api/v1',
   },
+  MAILTRAP: {
+    EMAIL_HOST: EMAIL_HOST,
+    EMAIL_USERNAME: EMAIL_USERNAME,
+    EMAIL_PASSWORD: EMAIL_PASSWORD,
+    FROM_EMAIL: FROM_EMAIL,
+  },
```
oleksandr-danylchenko, line 40: Shouldn't all these vars be under the `NODEMAILER` object? Because, as I understand, we use the `nodemailer` instead of the `mailtrap`. Which is totally cool, but I just want to be consistent :slightly_smiling_face:

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):
LGTM

PR #19, opened 2022-08-11
Fld 'user.photo_url' set as optional (may be NULL)

stanislav-ostr:


oleksandr-danylchenko:
Did you change the schema and migration without creating a new one?

stanislav-ostr:
> Did you change the schema and migration without creating a new one?

Yes, minor changes

stanislav-ostr:
Whould I create new migration?

oleksandr-danylchenko:
> Should I create new migration?

Yes. Prisma forces us to create new migrations even for small changes. And we mustn't modify the existing ones. Because Prisma will throw the existing migration conflict error for other devs and won't process the changes

oleksandr-danylchenko (commented):
Was the `yarn.lock` changed after the migration? Because I don't think that it should be included in this PR :slightly_smiling_face:

stanislav-ostr:
I don't understand how it happened, I just ran yarn install:all

stanislav-ostr:
ready

oleksandr-danylchenko (approved):
Thanks!

PR #21, opened 2022-08-11
Sign up logic for users

o-rudych:
Я використав функцію
`{ mailSend } from '@services/mail_verification/send.service';`
у гілці **develop** її немає, але я взяв з гілки **email_verification**.
Звідти взяв файли:
```
backend/src/services/mail_verification/send.service.ts
backend/src/services/mail_verification/token.service.ts
backend/src/services/mail_verification/user_security.service.ts

```


backend/src/common/enums/app/env.enum.ts
```diff
@@ -31,6 +31,16 @@ const ENV = {
...
   },
+  MAIL: {
+    API_URL: 'http://localhost:3000/',
+    ACTIVATE_URL: 'http://localhost:3001/api/v1/activate/',
+    SUCCESS_URL: 'http://localhost:3000/mail-success',
+    FAILED_URL: 'http://localhost:3000/mail-failed',
+    SMTP_HOST: 'smtp.gmail.com',
+    SMTP_PORT: 587,
+    SMTP_USER: 'autoline.binary@gmail.com',
+    SMTP_PASSWORD: 'pyvwijwvexspvzfi',
```
RomanMahola: Are you guys serious? 
### NEVER PROVIDE CREDENTIALS IN CODE!!!
Why did not use .env?

backend/src/middlewares/auth/auth.middleware.ts
```diff
@@ -32,4 +35,26 @@ const localAuth = (
...
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    await signUpSchema.validate(req.body);
+    const existedUser = await prisma.user.findFirst({
+      where: {
+        email: req.body.email,
+      },
+    });
+    if (existedUser) throw new Error();
```
RomanMahola: I don`t understand.
If user exists you throw new error?

backend/src/middlewares/auth/auth.middleware.ts
```diff
@@ -32,4 +35,26 @@ const localAuth = (
...
+
+    next();
+  } catch (err) {
+    console.error(err);
+    res.status(403);
```
RomanMahola: Use constants 
httpStatus.FORBIDDEN

backend/src/middlewares/auth/auth.middleware.ts
```diff
@@ -32,4 +35,26 @@ const localAuth = (
...
+    next(new Error('Some error'));
```
RomanMahola: Some error have no logic context. You should handle errors

backend/src/services/mail_verification/send.service.ts
```diff
@@ -0,0 +1,42 @@
...
+  host: ENV.MAIL.SMTP_HOST,
+  port: ENV.MAIL.SMTP_PORT,
+  secure: false,
+  auth: {
+    user: ENV.MAIL.SMTP_USER,
+    pass: ENV.MAIL.SMTP_PASSWORD,
+  },
+});
+
+async function sendActivationMail(to: string, link: string): Promise<void> {
```
RomanMahola: https://github.com/BinaryStudioAcademy/bsa-2022-autoline/pull/17#discussion_r943544259

backend/src/validation-schemas/sign-up/sign-up.validation-schema.ts
```diff
@@ -0,0 +1,12 @@
+import * as Yup from 'yup';
+
+const signUpSchema = Yup.object().shape({
+  name: Yup.string().required('Full name is required'),
```
RomanMahola: Also use max validation rule

For example 100 would be enough

RomanMahola (changes requested):

backend/src/middlewares/auth/auth.middleware.ts
oleksandr-danylchenko: @RomanMahola 
This is the sign **_up_** (registration) middleware. So throwing an exception is actually expected :slightly_smiling_face: . The only thing I would change is to add the message the `Error` constructor

backend/src/middlewares/auth/auth.middleware.ts
oleksandr-danylchenko: Yep, the hardcoded values mustn't be used in code. Or used only when it makes 100% sense to everyone else and it's just unavoidable under any circumstances

backend/src/common/enums/app/env.enum.ts
o-rudych: I agree, fixed it. I left it to Serhii who does email verification task.

backend/src/middlewares/auth/auth.middleware.ts
o-rudych: Yes, I added error message.

backend/src/middlewares/auth/auth.middleware.ts
o-rudych: Yes, thanks for the tip. Fixed it.

backend/src/middlewares/auth/auth.middleware.ts
o-rudych: I did this at first, because when Oleksandr said that it is better to always return a 403 error and so that the hacker does not understand the reason of failure, from a security point of view.
I fixed it, now it returns an error message.

backend/src/services/mail_verification/send.service.ts
o-rudych: I left it to Serhii, who does this task.

backend/src/validation-schemas/sign-up/sign-up.validation-schema.ts
o-rudych: Thanks for the tip. Added new rule.

oleksandr-danylchenko (approved):
Looks good!

PR #22, opened 2022-08-11
feat (docker) Add Docker file to mariaDB.

RomanMahola:
- Add Worskpaces
- Add docker-compose.yml


backend/.env.example
```diff
@@ -0,0 +1,32 @@
...
+MYSQL_ROOT_PASSWORD=secret
+MYSQL_PASSWORD=secret
+
+# Use these for the Docker development
+MYSQL_HOST=autoline_mariadb
+MYSQL_PORT=3306
+
+# Use these for the local development
+#MYSQL_HOST=localhost
+#MYSQL_PORT=3307
```
oleksandr-danylchenko: As we don't use Docker for the dev, let's uncomment the second `MYSQL` vars option

package.json
```diff
@@ -6,6 +6,13 @@
     "yarn": ">=1 <2"
   },
   "scripts": {
+    "start:frontend": "npm run start -w frontend",
+    "start:backend": "npm run build -w shared && npm run start:dev -w backend",
+    "migrate": "npm run migrate -w backend",
```
oleksandr-danylchenko: Please, don't use both the `npm` & `yarn` in the same `package.json` :pray: 

--- 

`shared` has its own `start:dev` script. Also, we shouldn't run the shared package as part of some other app start script

```suggestion
    "start:frontend": "yarn start -w frontend",
    "start:backend": "yarn start:dev -w backend",
    "start:shared": "yarn start:dev -w shared",
```

.docker/autoline/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+    hostname: mariadb
+    ports:
+      - "3307:3306"
+    volumes:
+      - autoline_mariadb_data:/var/lib/mysql
+    environment:
+      - "MYSQL_DATABASE=autoline"
+      - "MYSQL_USER=autoline"
+      - "MYSQL_PASSWORD=secret"
+      - "MYSQL_ROOT_PASSWORD=secret"
```
oleksandr-danylchenko: It's better to include the backend env file than to use the inline variables. In that way, there will be single source of truth for the DB connection and there will be no possibility to use wrong credentials in the backend

oleksandr-danylchenko (changes requested):
It looks simple enough but some polishing is required

.docker/autoline/docker-compose.stage.yml
```diff
@@ -0,0 +1,59 @@
...
+    restart: on-failure
+    ports:
+      - "3306:3306"
+    volumes:
+      - autoline_mariadb_data:/var/lib/mysql
+    env_file:
+      - ../../.env/db.env
+    networks:
+      - autoline_backend_network
+  autoline_backend:
```
oleksandr-danylchenko: A bit of space would be awesome
```suggestion
      - autoline_backend_network

  autoline_backend:
```

.docker/autoline/docker-compose.yml
```diff
@@ -1,25 +1,20 @@
-version: "3.8"
-
+version: '3'
 services:
   autoline_mariadb:
-    container_name: autoline_mariadb
+    container_name: 'autoline_mariadb'
```
oleksandr-danylchenko, line 4: quotes are not needed

.docker/autoline/docker-compose.yml
```diff
@@ -1,25 +1,20 @@
...
 
 volumes:
   autoline_mariadb_data:
     driver: local
-
-networks:
-  autoline_backend_network:
-    driver: bridge
+  shared:
+    driver: local
```
oleksandr-danylchenko, line 20: Why do we need the `shared` volume here?

.docker/autoline/docker-compose.yml
```diff
@@ -1,25 +1,20 @@
...
 services:
   autoline_mariadb:
-    container_name: autoline_mariadb
+    container_name: 'autoline_mariadb'
     build:
-      context: ../../
+      context: ./
       dockerfile: database/Dockerfile
-    restart: on-failure
+    hostname: mariadb
```
oleksandr-danylchenko, line 8: Whoa, how the `mariadb` `hostname` works?

PR #23, opened 2022-08-11
fix (structure) Fix structure of backend

RomanMahola:
- Add .env.example to backend directory
- Change configs


backend/.env.example
```diff
@@ -1,32 +1,44 @@
...
+#####################################
 # Use these for the Docker development
+#####################################
+
 MYSQL_HOST=autoline_mariadb
 MYSQL_PORT=3306
 
-# Use these for the local development
-#MYSQL_HOST=localhost
-#MYSQL_PORT=3307
```
oleksandr-danylchenko: These variables **mustn't** be removed! From the outer world, we can connect only to the exposed port, which is 3307 and `localhost`. `autoline_mariadb` and `3306` exists only in the Docker-Compose environment and cannot be addressed

oleksandr-danylchenko (changes requested):
We need to get the local MYSQL vars back)

PR #24, opened 2022-08-11
fix (structure) Fix structure of frontend

RomanMahola:
- Add .env.example to frontend directory
- Change configs


frontend/.env.example
```diff
@@ -0,0 +1,3 @@
+VITE_API_ORIGIN_URL=http://localhost:3001/api/v1
+VITE_SENTRY_DSN='your sentry dsn'
+SENTRY_ENABLE=false
```
oleksandr-danylchenko: `SENTRY_ENABLE` also needs to have the `VITE_` prefix. W/o it, the variable won't be recognized
https://vitejs.dev/guide/env-and-mode.html#env-files

oleksandr-danylchenko: We also miss the `VITE_SENTRY_DSN`

frontend/src/configs/config.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export interface ConfigInterface {
+  APP: AppConfig;
+}
+
+const configuration = (): ConfigInterface => {
+  config();
+
+  const { VITE_API_ORIGIN_URL, VITE_SENTRY_DSN, SENTRY_ENABLE } = process.env;
```
oleksandr-danylchenko: We **cannot** import env variables from the `process.env` in Vite. The variables must be imported from the `import.meta.env`

oleksandr-danylchenko (changes requested):
Needs some Vite-specific polishing

frontend/src/configs/config.ts
RomanMahola: Do we need additional config to import.meta.env. I mean path to .env file

oleksandr-danylchenko: No, if the env file is in the same directory as the `package.json`. Actually... we don't even need to use the `dotenv` explicitly by ourselves, bc `vite` uses it under the hood already

PR #25, opened 2022-08-11
Fix Workspaces

RomanMahola:


package.json
```diff
@@ -6,25 +6,25 @@
...
   },
   "scripts": {
-    "start:frontend": "npm run start:dev -w frontend",
-    "start:backend": "npm run build -w shared && npm run start:dev -w backend",
-    "migrate": "npm run migrate -w backend",
-    "install:frontend": "yarn -w frontend",
-    "install:backend": "yarn -w backend",
-    "install:shared": "yarn -w shared",
+    "start:frontend": "yarn workspace @autoline/frontend run start:dev:local",
+    "start:backend": "yarn workspace @autoline/shared build && yarn workspace @autoline/backend start:dev",
```
oleksandr-danylchenko: The same thing as in the https://github.com/BinaryStudioAcademy/bsa-2022-autoline/pull/22/files#r943447620
We shouldn't build the shared package only for backend, it has its own `start:dev`

oleksandr-danylchenko (changes requested):

PR #26, opened 2022-08-11
GitHub actions

RomanMahola:


.docker/autoline/_docker-compose.dev.yml
```diff
@@ -0,0 +1,76 @@
...
+
+volumes:
+  autoline_mariadb_data:
+    driver: local
+
+networks:
+  autoline_backend_network:
+    driver: bridge
+  autoline_frontend_network:
+    driver: bridge
```
oleksandr-danylchenko: Do we need this file? Does it work on Windows?

.docker/autoline/backend.Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM node:16.16.0-slim as backend-build
+
+RUN apt-get update
+RUN apt-get install -y openssl
+
+WORKDIR /app
+
+COPY ./package*.json ./
```
oleksandr-danylchenko: We need to copy `yarn.lock`
```suggestion
COPY ./package.json ./
COPY ./yarn.lock ./
```

.docker/autoline/docker-compose.dev.yml
```diff
@@ -1,74 +1,54 @@
 version: "3.8"
 
-# Using `node_modules` mounting pattern described here:
-# https://burnedikt.com/dockerized-node-development-and-mounting-node-volumes/#prevent-hosts-node_modules-from-being-mounted
-
 services:
   autoline_mariadb:
+    image: mariadb
```
oleksandr-danylchenko: Prisma migration **won't work** on the sole `mariadb` image. Because the default image allows its user to access only one database. Prisma requires to have access to all the DBs in the system. So we must have the `user_priviligies.sql` script added here

.docker/autoline/docker-compose.dev.yml
```diff
@@ -1,74 +1,54 @@
 version: "3.8"
```
oleksandr-danylchenko: Why is this `docker-compose` called `.dev`? There's no reloading of the code on change. We even use the remote images

.docker/autoline/docker-compose.dev.yml
```diff
@@ -1,74 +1,54 @@
...
+    image: mariadb
     container_name: autoline_mariadb
     build:
-      context: ../../
+      context: ./
       dockerfile: database/Dockerfile
     restart: on-failure
     ports:
-      - "3307:3306"
+      - "3306:3306"
```
oleksandr-danylchenko: The `3307` is specifically chosen to not conflict with the MySQL database which can be running locally

.docker/autoline/frontend.Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM node:16.16.0-alpine as frontend-build
+
+WORKDIR /app
+
+COPY ./package*.json ./
```
oleksandr-danylchenko: We need to copy `yarn.lock`
```suggestion
COPY ./package.json ./
COPY ./yarn.lock ./
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/checkout@v2
+
+      - name: Install NodeJS
+        uses: actions/setup-node@v2
+        with:
+          node-version-file: .nvmrc
+
+      - name: Install Dependencies
+        run: |
+          yarn install
```
oleksandr-danylchenko, line 25: Please use the dependencies install cache. It will save us hours

Here's the action which I used initially - https://github.com/whoan/docker-build-with-cache-action

oleksandr-danylchenko (changes requested):

.github/workflows/ci.yml
oleksandr-danylchenko: Please use the dependencies install cache. It will save us hours

Here's the action which I used initially - https://github.com/bahmutov/npm-install

```suggestion
      - name: Install Dependencies
         uses: bahmutov/npm-install@v1.6.0
         with:
           useRollingCache: true
```

.docker/autoline/_docker-compose.dev.yml
RomanMahola: No. This is old version of the file. I did not mention that and committed it, sorry.

.docker/autoline/backend.Dockerfile
RomanMahola: Got you

.docker/autoline/docker-compose.dev.yml
RomanMahola: dev because of dev server. If we want up docker localy we can just leave docker-compose.yml or docker-compose.local

.docker/autoline/docker-compose.dev.yml
RomanMahola: No because it is configuration for staging.

PR #27, opened 2022-08-11
Add redux toolkit integration

brovarenko:
See - https://app.asana.com/0/1202525679663009/1202782033632765/f


frontend/src/store/store.ts
```diff
@@ -0,0 +1,7 @@
+import { configureStore } from '@reduxjs/toolkit';
+
+import { rootReducer } from './root-reducer';
+
+export const store = configureStore({
+  reducer: rootReducer,
+});
```
oleksandr-danylchenko: It's a better practice to keep the `combineReducers` (declared in `frontend/src/store/root-reducer.ts` now) in the same file before the `configureStore`. It'll make store files reading easier

frontend/src/store/auth/slice.ts
```diff
@@ -0,0 +1,17 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+type State = {
+  user: null;
+};
+
+const initialState: State = {
+  user: null,
+};
```
oleksandr-danylchenko: It's better to call `State` more specific - `AuthState`

frontend/src/main.tsx
```diff
@@ -1,10 +1,13 @@
 import React from 'react';
 import { createRoot } from 'react-dom/client';
+import { Provider } from 'react-redux';
 
 import { ENV } from '@common/enums/app/env.enum';
 import { Routing } from '@navigation/routing/routing';
 import * as Sentry from '@sentry/react';
 import { BrowserTracing } from '@sentry/tracing';
+
+import { store } from './store/store';
```
oleksandr-danylchenko: Could you please replace all the relative `/store` imports with the `@store` alias?

oleksandr-danylchenko (changes requested):
Great work! Just a bit of the final touches required

oleksandr-danylchenko (approved):

PR #28, opened 2022-08-11
Add roles-based protection

R0ixy:
https://app.asana.com/0/1202525679663009/1202720801120408/f

### DIFF
--- a/backend/src/middlewares/roles-based-protection/role.middleware.ts
+++ b/backend/src/middlewares/roles-based-protection/role.middleware.ts
@@ -1,0 +1,36 @@
+import { TokenPayload } from '@autoline/shared';
+import { ExceptionMessage } from '@common/enums/exception/exception-message.enum';
+import { TypedRequestBody } from '@common/types/controller/controller';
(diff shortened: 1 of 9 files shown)


backend/src/middlewares/roles-based-protection/role.middleware.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    if (token) {
+      token = token.replace('Bearer ', '');
+      const payload = verifyToken(token) as TokenPayload;
+
+      if (roles.includes(payload.role as Role)) {
+        req.body.tokenPayload = payload;
+        next();
+        return;
+      }
```
oleksandr-danylchenko: This can be shortened to the:
```suggestion
      if (roles.includes(payload.role as Role)) {
        req.body.tokenPayload = payload;
        return next();
      }
```

oleksandr-danylchenko (dismissed):
All looks good! Thanks!

oleksandr-danylchenko (approved):

PR #29, opened 2022-08-12
Add wishlist - backend part

iryb:
Task: https://app.asana.com/0/1202525679663009/1202720801120417

## Added:

- User_Wishlist model and migration
- Routes `/wishlist/model/:id` and `/wishlist/complectation/:id`
- Wishlist controller
- `setWishlist` function in service to delete or create wishlist

## Issue:

- [ ] `http://localhost:3001/wishlist/model/3deba182-4766-4aa4-8a91-2478282dfa2e` POST request with `user_id` key in body returns `Cannot POST`. Where `3deba182-4766-4aa4-8a91-2478282dfa2e` is random id.

## Questions:

- [x] 1. Should I implement `getUserWishlist` function to wishlist service? This function will return list of wishlists based on user id.
- [x] 2. For frontend part I am going to only create wishlist service. We don't have Redux integrated yet. Is that what was expected for frontend part of this task?


iryb:
@RomanMahola

iryb:
Okay, I found the issue and route works now. I am testing the functionality.

backend/prisma/schema.prisma
```diff
@@ -72,6 +82,7 @@ model Model {
   autoria_code           Int
   complectations         Complectation[]
   prices_ranges          Prices_Range[]
+  users_wishlist         User_Wishlist[]
```
oleksandr-danylchenko: I would prefer if both words are in the plural form: `users_wishlist` -> `users_wishlists`

backend/prisma/schema.prisma
```diff
@@ -127,6 +138,7 @@ model Complectation {
   transmission_type    Transmission_Type         @relation(fields: [transmission_type_id], references: [id])
   options              ComplectationsOnOptions[]
   prices_ranges        Prices_Range[]
+  users_wishlist       User_Wishlist[]
```
oleksandr-danylchenko: `users_wishlist` -> `users_wishlists`

backend/src/routes/preferences/wishlist.router.ts
```diff
@@ -0,0 +1,15 @@
...
+const PATH = '/wishlist';
+
+const wishlistRouter = Router();
+
+wishlistRouter.post(`${PATH}/model/:id`, wishlistController.setModelWishlist);
+
+wishlistRouter.post(
+  `${PATH}/complectation/:id`,
+  wishlistController.setComplectationWishlist,
+);
```
oleksandr-danylchenko: These routes can be **simplified in one simple change**. Plus, there will be no duplication of the code in the endpoints. Actually, there will be **only one endpoint**.

You need to use a bit different route pattern: 
```suggestion
const wishlistRouter = Router();

wishlistRouter.post(
  `${PATH}/wishlist/:userId`,
  wishlistController.setWishlist,
);
```
And then you can provide either the `?modelId=` or the `?complectationId=` as the URL params.
Example: `http://localhost:3001/someUserId/wishlist?modelId=`

Then, in the `setWishlist` you can do smth like:
```ts
    ...
    const { userId } = req.body;
    const { modelId, complectationId } = req.params;

    const wishlist = {
      userId,
      modelId,
      complectationId,
    };

    ...
```

This way, the whole wishlist logic takes much less lines of code and adding the new wishlist type would be very easy

backend/src/services/preferences/wishlist.service.ts
```diff
@@ -0,0 +1,43 @@
+import { prisma } from '@data/prisma-client';
+
+import type { WishlistResponseDto } from '@autoline/shared';
+import type { WishlistInput } from '@common/types/types';
+
+const setWishlist = async (
+  input: WishlistInput,
+): Promise<WishlistResponseDto> => {
+  const { user_id, model_id, complectation_id } = input;
```
oleksandr-danylchenko: We **must** use only **`camelCase`** names for the variables in the JS code. Please, don't use `snake_case` when it's possible

backend/src/common/types/prisma/prisma.ts
```diff
@@ -4,4 +4,10 @@ interface UserCreateInput extends Prisma.UserCreateInput {
   password: string;
 }
 
-export type { UserCreateInput };
+interface WishlistInput {
+  user_id: string;
+  model_id?: string;
+  complectation_id?: string;
+}
```
oleksandr-danylchenko: Please don't use `snake_case`

backend/src/services/preferences/wishlist.service.ts
```diff
@@ -0,0 +1,43 @@
...
+    },
+  });
+
+  const { id: newWishlistId } = wishlist
+    ? await prisma.user_Wishlist.delete({
+        where: {
+          id: wishlist.id,
+        },
+      })
+    : await prisma.user_Wishlist.create({
```
oleksandr-danylchenko: Didn't Prisma generate the `User_Wishlist` with upper _U_ at the start?

oleksandr-danylchenko (changes requested):
Some optimization and simplification are required. But, in any way, I like your attempt :raised_hands:

oleksandr-danylchenko:
And could you please resolve the conflicts?

backend/src/services/preferences/wishlist.service.ts
iryb: @oleksandr-danylchenko I can't find out how to fix this. Prisma generates model with first letter in lower case `user_Wishlist`. Other users faced this issue too https://github.com/prisma/prisma/issues/5035

oleksandr-danylchenko: Ohhh... I got this. Prisma, by default, uses the `camelCase` for the tables. So it just uses the lowercase for all the generated tables - https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/use-custom-model-and-field-names#:~:text=Prisma%20has%20a%20naming%20convention%20of%20camelCasing

backend/src/common/types/prisma/prisma.ts
```diff
@@ -4,4 +4,11 @@ interface UserCreateInput extends Prisma.UserCreateInput {
   password: string;
 }
 
-export type { UserCreateInput };
+interface WishlistInput {
+  [key: string]: string | undefined;
```
oleksandr-danylchenko: Did you add the `[key: string]: string | undefined;` because there was an error when you tried to use it as the generic? The `type` should be used instead of the interface in that case.
Because `[key: string]: string | undefined;` makes you interface extremely loosely typed

backend/src/services/preferences/wishlist.service.ts
```diff
@@ -0,0 +1,43 @@
...
+      complectation_id: complectationId,
+    },
+  });
+
+  const { id: newWishlistId } = wishlist
+    ? await prisma.user_Wishlist.delete({
+        where: {
+          id: wishlist.id,
+        },
+      })
```
oleksandr-danylchenko: I don't think we should delete a wishlist record in the POST request. It should be done in a separate DELETE request. In my view, if the document for such a model and the user exists, we can return _400 BAD REQUEST_ error

oleksandr-danylchenko (changes requested):

backend/src/controllers/preferences/wishlist.controller.ts
```diff
@@ -25,8 +25,37 @@ const setWishlist = async (
     res.json(wishlistResponseDto).status(httpStatus.CREATED);
   } catch (error) {
     console.error(error);
+    if (error instanceof Error) {
+      res.statusCode = 400;
+    }
```
oleksandr-danylchenko: Please use `res.sendStatus()` method - https://flaviocopes.com/express-send-response/

backend/src/controllers/preferences/wishlist.controller.ts
```diff
@@ -25,8 +25,37 @@ const setWishlist = async (
...
+      complectationId,
+    };
+
+    await wishlistService.deleteWishlist(wishlist);
+    res.json('Wishlist deleted successfully').status(httpStatus.CREATED);
+  } catch (error) {
+    console.error(error);
+    if (error instanceof Error) {
+      res.statusCode = 400;
+    }
```
oleksandr-danylchenko: Use `sendStatus` + the status should be 404

backend/src/services/preferences/wishlist.service.ts
```diff
@@ -40,4 +38,28 @@ const setWishlist = async (
...
+    where: {
+      user_id: userId,
+      model_id: modelId,
+      complectation_id: complectationId,
+    },
+  });
+
+  if (!wishlist) {
+    throw new Error('Bad request');
+  }
```
oleksandr-danylchenko: Message should be more specific and tied to the service method logic

oleksandr-danylchenko (changes requested):
You're getting closer! Just a few small fixes left :raised_hands:

oleksandr-danylchenko (approved):

PR #30, opened 2022-08-12
Added header

AlexanderKhapchenko:
See https://app.asana.com/0/1202525679663009/1202720798232758/f

Viewing without user authentication
Знімок екрана з 2022-08-12 15-31-14

Viewing an authenticated user
Знімок екрана з 2022-08-12 15-31-42

Viewing an authenticated user with an open menu
Знімок екрана з 2022-08-12 15-31-56


frontend/src/components/header/header.module.scss
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.header {
+  position: relative;
+  z-index: 2;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 100%;
+  height: 105px;
```
oleksandr-danylchenko: Usually, the height of the header is considered as the project CSS variable. Because it can be reused in multiple places to properly calculate the size or the offsets. So I would recommend to put it in the root of the styles and then use as CSS variable

frontend/src/components/header/private-elements/private-elements.module.scss
```diff
@@ -0,0 +1,70 @@
...
+    li {
+      display: flex;
+      justify-content: center;
+      width: 100%;
+      padding: 5px 0;
+    }
+
+    li:hover {
+      background: #ececec;
+    }
```
oleksandr-danylchenko: That's useful that you invested time into developing your own popover window. But I think that we can use the [built-in MUI solution](https://mui.com/material-ui/react-popover/) for this use-case

frontend/src/components/header/private-elements/private-elements.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import ArrowDown from '@assets/images/header/arrow-drop-down.svg';
+import DefaultAvatar from '@assets/images/header/default-avatar.png';
+
+import { IRemindersProps, Reminders } from '../reminders/reminders';
+import styles from './private-elements.module.scss';
+
+interface IPrivateComponentProps extends IRemindersProps {
+  avatar: string | undefined | null;
+}
```
oleksandr-danylchenko: Please don't use `I` in front of the interfaces. It's not considered as the best practice in TS

frontend/src/components/header/private-elements/private-elements.tsx
```diff
@@ -0,0 +1,53 @@
...
+        comparisons={comparisons}
+        notifications={notifications}
+      />
+      <div
+        className={styles.avatarCover}
+        onClick={(): void => setIsMenuOpen(!isMenuOpen)}
+      >
+        <img
+          className={styles.avatar}
+          src={avatar ? avatar : DefaultAvatar}
```
oleksandr-danylchenko: Statements that look like `smth ? smth : smth_else` must be replaced with the `smth || smth_else`

frontend/src/components/header/private-elements/private-elements.tsx
```diff
@@ -0,0 +1,53 @@
...
+          alt="avatar"
+        />
+        <img
+          className={`${styles.arrow} ${isMenuOpen ? styles.active : ''}`}
```
oleksandr-danylchenko: Use can use utility called [clsx](https://www.npmjs.com/package/clsx) to conditionally combine classnames w/o using the `statement ? styles : ''`

frontend/src/components/header/private-elements/private-elements.tsx
```diff
@@ -0,0 +1,53 @@
...
+    <div className={styles.information}>
+      <Reminders
+        favorites={favorites}
+        comparisons={comparisons}
+        notifications={notifications}
+      />
+      <div
+        className={styles.avatarCover}
+        onClick={(): void => setIsMenuOpen(!isMenuOpen)}
+      >
```
oleksandr-danylchenko: It's not semantically correct to apply any action listeners to the non-interactive elements. Such listeners need to be applied to the `button`s, links and other native clickable elements. Please consider using the MUI's [IconButton](https://mui.com/material-ui/react-button/#icon-button) & [Avatar](https://mui.com/material-ui/react-avatar/#main-content)

frontend/src/components/header/reminders/reminders.module.scss
```diff
@@ -0,0 +1,36 @@
...
+    align-items: center;
+    width: 15px;
+    height: 15px;
+    background-color: #3765ff;
+    border-radius: 50%;
+
+    span {
+      color: #ffffff;
+      font-size: 10px;
+    }
```
oleksandr-danylchenko: You could use the MUI's [`Badge`](https://mui.com/material-ui/react-badge/) component. But that's also good that you now know how to create badges in CSS :ok_hand:

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (dismissed):

oleksandr-danylchenko (approved):

PR #39, opened 2022-08-16
Create an images carousel for the detailed view

ashonuyak:
See - https://app.asana.com/0/1202525679663009/1202804889720777/f
https://gyazo.com/614f5369eef263569205663a2fcb6c00


oleksandr-danylchenko (changes requested):
Please create a separate PR that will include **only** the carousel changes. Also, for visual PRs screenshots or videos are needed to evaluate the correctness

oleksandr-danylchenko (commented):
Seeing some screenshots would be highly appreciated. W/o I cannot image what I'm about to merge :sweat_smile:

frontend/src/components/common/carousel/styles.css
```diff
@@ -0,0 +1,90 @@
...
+  object-fit: cover;
+}
+
+.mySwiper2 {
+  width: 538px;
+  height: 376px;
+}
+.mySwiper {
+  height: 100px;
+  width: 538px;
```
nikita-remeslov: Round this value and if you use it more then 1 time, move it to variable

frontend/src/components/common/carousel/styles.css
```diff
@@ -0,0 +1,90 @@
...
+  bottom: 0;
+  height: 3px;
+  width: 98%;
+  background-color: blue;
+  border-bottom-left-radius: 30%;
+  border-bottom-right-radius: 30%;
+}
+
+.mySwiper2 {
+  border-radius: 10px;
```
nikita-remeslov: Do we have default variables for border-radius, padding, etc.?

frontend/src/components/common/carousel/styles.css
```diff
@@ -0,0 +1,90 @@
...
+  width: 90px;
+  height: 40px;
+  bottom: 30px;
+  left: 30px;
+  border-radius: 3.24px;
+}
+.navigation-bar .current-image {
+  font-size: 16px;
+  font-weight: 400;
+  font: Lato, Arial, sans-serif;
```
nikita-remeslov: We should have variable for font

frontend/src/components/common/carousel/carousel.tsx
```diff
@@ -0,0 +1,84 @@
...
+import 'swiper/css/thumbs';
+import IconPark from './icon-park-solid_picture.svg';
+
+import type { Swiper as SwiperType } from 'swiper';
+
+export const Carousel = (): React.ReactElement => {
+  const [thumbsSwiper, setThumbsSwiper] = useState<SwiperType | null>(null);
+  const [currentImage, setCurrentImage] = useState(1);
+
+  const images = [
```
nikita-remeslov: Move it to a different file, because this component should use props to render the image

And if you try to pass images to this component, nothing will work

frontend/src/components/common/carousel/carousel.tsx
```diff
@@ -0,0 +1,84 @@
...
+import React, { useState } from 'react';
+
+import { FreeMode, Navigation, Thumbs } from 'swiper';
+import { Swiper, SwiperSlide } from 'swiper/react';
+
+import 'swiper/css';
+import 'swiper/css/free-mode';
+import 'swiper/css/navigation';
+import 'swiper/css/thumbs';
+import IconPark from './icon-park-solid_picture.svg';
```
nikita-remeslov: we should move to asset folder this icon

frontend/src/components/common/carousel/styles.css
```diff
@@ -0,0 +1,90 @@
...
+.mySwiper2:hover .swiper-button-prev::after {
+  margin-left: 40px;
+}
+
+.navigation-bar {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  position: absolute;
+  z-index: 999;
```
nikita-remeslov: bad practice use so big value for z-index. In the perfect world, we should define variables for z-index and use it

I prefer to do something like this but with smaller numbers - https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L1043-L1052

nikita-remeslov (changes requested):
Is it possible to make 5 pictures under the main picture? Since it's done in design

frontend/src/components/common/carousel/styles.css
ashonuyak: i took almost all variables from figma

nikita-remeslov:
@ashonuyak also we use scss in the project, remove all CSS files

frontend/src/components/common/carousel/styles.scss
```diff
@@ -0,0 +1,102 @@
+$swiper-width: 540px;
+$font-lato: lato, arial, sans-serif;
+$z-index-current-image: 1;
+$preview-image-height: 70px;
+$main-image-height: 380px;
```
oleksandr-danylchenko: Please use native CSS variables. Because we already use them to define root constants. We shouldn't use both the SCSS and CSS variables at the same time

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #40, opened 2022-08-17
Add general components

iryb:
Added common components that will be used on all pages:

- Title
- Page container

Added route for Personal page and `PersonalPage` component.

Fixed `types` files extensions (.tsx to .ts)


frontend/src/assets/css/typography.scss
```diff
@@ -0,0 +1,22 @@
+h3 {
+  font-size: 58px;
```
nikita-remeslov: Do we have a default value for font-size? 
I think it is better to use **rem**. Because in this case if we need to update font-size we will need to update all px values

frontend/src/components/common/page-container/styles.module.scss
```diff
@@ -0,0 +1,5 @@
+.container {
+  &.containerLarge {
+    max-width: 1670px;
```
nikita-remeslov: If I remember the daily discussion we decided to use the default lib container size

nikita-remeslov (changes requested):

frontend/src/assets/css/typography.scss
iryb: @nikita-remeslov @oleksandr-danylchenko liner doesn't allow me to use `rem`: `Unexpected unit "rem"  unit-disallowed-list`

iryb: Someone added font sizes in `%`. Is this way suitable for current case too?

iryb:
@nikita-remeslov I can not understand why there are so many files with deletions (about 49 files changed)

iryb:
> 
okay, it goes from `Resolve conflicts` commit. Probably it will break something on dev branch, no?

oleksandr-danylchenko (dismissed):

PR #41, opened 2022-08-17
Fixed Node-exclusive `shared` package complication

oleksandr-danylchenko:
Previously the `shared` package was built to work with the `commonjs` modules: 
image
 And `Vite` doesn't support that. So the `frontend` couldn't import JS code from the `shared` package

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -1,8 +1,9 @@
 /* eslint-disable import/no-default-export */
 
-import { defineConfig } from 'vite';
-import react from '@vitejs/plugin-react';
 import path from 'path';
+
+import react from '@vitejs/plugin-react';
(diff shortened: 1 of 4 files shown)


frontend/vite.config.ts
```diff
@@ -25,6 +26,13 @@ export default defineConfig({
...
       '@navigation': path.resolve(__dirname, './src/navigation'),
+      '@autoline/shared': path.resolve(
+        __dirname,
+        './node_modules/@autoline/shared',
+      ),
     },
   },
+  optimizeDeps: {
+    exclude: ['@autoline/shared'],
+  },
```
oleksandr-danylchenko, line 37: Vite won't cache the local lib

nikita-remeslov (approved):
Looks good

PR #42, opened 2022-08-17
Add cars filtering

R0ixy:
Asana - https://app.asana.com/0/1202525679663009/1202720801120415/f


backend/src/helpers/cars/api_autoria.helper.ts
```diff
@@ -0,0 +1,23 @@
+import { ENV } from '@common/enums/app/env.enum';
+import { AutoriaRequestParams } from '@common/types/types';
+import axios from 'axios';
+
+const URL = 'https://developers.ria.com/auto/search';
```
nikita-remeslov: Maybe better move it to env? @oleksandr-danylchenko

backend/src/common/types/cars/autoria_request_params.ts
```diff
@@ -0,0 +1,12 @@
+interface AutoriaRequestParams {
+  category_id: number;
+  price_ot?: string;
+  price_do?: string;
+  powerFrom?: string;
+  powerTo?: string;
+  power_name?: number;
+  engineVolumeFrom?: string;
+  engineVolumeTo?: string;
```
nikita-remeslov: Are you sure you need that many optional keys?
In your case for request, we can use `Partial<AutoriaRequestParams>`

backend/src/helpers/cars/api_autoria.helper.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const getCarsAutoRia = async (
+  requestParams: AutoriaRequestParams,
```
nikita-remeslov:
```suggestion
  requestParams: Partial<AutoriaRequestParams>,
```

nikita-remeslov (changes requested):

backend/src/helpers/cars/api_autoria.helper.ts
oleksandr-danylchenko: I think it's okay to leave it in the source code. It won't change depending on ENV and I have high hopes that the autoria api URL won't change in a next month

backend/src/common/types/cars/autoria_request_params.ts
oleksandr-danylchenko: :arrow_double_up: :+1:

backend/src/controllers/cars/cars.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+const carsSearch = async (
+  req: TypedRequestQuery<CarsSearchParams>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const carsData = await carsService.carsSearch(req.query);
+    const AutoRiaCarsData = await getCarsAutoRia(carsData);
+
+    res.json(AutoRiaCarsData);
```
oleksandr-danylchenko: Please use `camelCase`

backend/src/controllers/cars/cars.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+  } catch (error) {
+    if ((error as Error)?.message === 'NotFoundError') {
+      res.json([]);
+    } else {
+      next(error);
+    }
```
oleksandr-danylchenko: You should use `if (error instanceof Error)` check

backend/src/services/cars/carsSearch.service.ts
```diff
@@ -0,0 +1,135 @@
+import { CarsSearchParams } from '@autoline/shared';
```
oleksandr-danylchenko: All the files should be named in `kebab-case`. Please rename the files named in `snake_case` and `camelCase`

backend/src/services/cars/carsSearch.service.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    model_from_db.forEach((model, index) => {
+      Object.assign(result, {
+        [`body_type[${index}]`]: model.body_type?.autoria_code,
+        [`marka_id[${index}]`]: model.brand?.autoria_code,
+        [`model_id[${index}]`]: data.modelId ? model.autoria_code : undefined,
+        [`s_yers[${index}]`]: data.yearStart,
+        [`po_yers[${index}]`]: data.yearEnd,
+      });
+    });
```
oleksandr-danylchenko: Could you explain what it does?

backend/src/services/cars/carsSearch.service.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  await getData(data.colorId, 'color_id', prisma.color, result);
+  await getData(
+    data.transmissionTypeId,
+    'gear_id',
+    prisma.transmission_Type,
+    result,
+  );
+  await getData(data.fuelTypeId, 'type_id', prisma.fuel_Type, result);
+  await getData(data.drivetrainId, 'drive_id', prisma.drivetrain, result);
```
oleksandr-danylchenko: Why the method which is named `getSmth` doesn't return the value? Mutating the parameter is a cursed anti-pattern straight from 1998

oleksandr-danylchenko (changes requested):
Some simplification and comments would be appreciated

backend/src/services/cars/carsSearch.service.ts
R0ixy: It's just the way autoria asks to specify params. If you want to search multiply models or brands you need to add them this way `marka_id[0]=423&marka_id[1]=222` and so on. And if you want to search by years you need to duplicate it for every model/brand. That's actually what this code does

oleksandr-danylchenko (approved):

PR #44, opened 2022-08-17
Added name to the `Complectation` table

Aleksandr-Shulha-dev:


oleksandr-danylchenko:
@Aleksandr-Shulha-dev 
This PR should be named more explicitly. Because it's not the fix, in the first place)))0)

oleksandr-danylchenko (approved):

PR #45, opened 2022-08-17
add seeders for cities and regions

R0ixy:
Asana - https://app.asana.com/0/1202525679663009/1202812994589249/f

### DIFF
--- a/backend/prisma/seed.ts
+++ b/backend/prisma/seed.ts
@@ -8,6 +8,8 @@ import {
   drivetrains,
   fuelTypes,
   transmissionTypes,
+  regions,
+  cities,
 } from '../api-autoria/cars/fetched-data/fetched-data';
 import { optionsTypes } from '../api-autoria/cars/options-types';
 import { OptionType } from '../api-autoria/option-type.enum';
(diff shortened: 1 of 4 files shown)


backend/prisma/seed.ts
```diff
@@ -28,6 +30,25 @@ async function main(): Promise<void> {
...
+  for (const region of regions) {
+    const dataToSeed = new AutoriaPlainDataDto(region);
+    const newRegion = await prisma.region.create({
+      data: dataToSeed,
+    });
+    const city = cities.filter(
+      (city: { data: { name: string; value: number }[]; regionId: number }) =>
+        city.regionId === region.value,
+    );
+    for (const data of city[0].data) {
```
oleksandr-danylchenko: There's a method `.find()` in JS which returns the first element that matched the search condition

```suggestion
    const city = cities.find(
      (city: { data: { name: string; value: number }[]; regionId: number }) =>
        city.regionId === region.value,
    );
    for (const data of city.data) {
```

oleksandr-danylchenko (changes requested):

backend/prisma/seed.ts
R0ixy: Ok, but then we need to cast type or check if `city` exists. Otherwise it will throw `TypeError: Object is possibly 'undefined'.`

oleksandr-danylchenko (approved):

PR #46, opened 2022-08-18
Roles management by admin on frontend

o-rudych:
I developed Admin panel for managing users on frontend on the route /administration.


frontend/src/components/administration/administration.tsx
```diff
@@ -0,0 +1,26 @@
...
+        'Loading data...'
+      ) : isSuccess ? (
+        data ? (
+          <UsersList users={data as User[]} />
+        ) : (
+          'User list is empty'
+        )
+      ) : (
+        'Error getting data'
+      )}
```
oleksandr-danylchenko: Please rewrite it to avoid the nesting of the ternary operators. Having ternary operator which invokes another ternary operator is considered as a bad practice

frontend/src/components/administration/users/dialog-edit-user.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const DialogEditUser: FC<DialogEditUserProps> = (props) => {
+  const { user, open, handleClose, handleSubmit } = props;
+  const [role, setRole] = useState<Role | undefined>(user?.role);
+
+  useEffect(() => {
+    if (user && open) {
+      setRole(user.role);
+    }
+  }, [user, open]);
```
oleksandr-danylchenko:
```suggestion
  useEffect(() => {
    if (user?.role && open) {
      setRole(user.role);
    }
  }, [user?.role, open]);
```
In such way minimize the rerenders, because we react to the primitive value instead of the object

backend/src/routes/user/user.router.ts
```diff
@@ -0,0 +1,34 @@
+import { Router } from 'express';
+
+const PATH = '/users';
+
+const userRouter = Router();
+
```
nikita-remeslov, line 6: Leave a comment that we need to remove this mock when we create this endpoint.

frontend/src/components/administration/users/users-list.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { useUpdateUserMutation } from '@store/queries/users';
+
+import { DialogEditUser } from './dialog-edit-user';
+
+type UsersListProps = {
+  users: User[];
+};
+
+const UsersList: FC<UsersListProps> = ({ users }) => {
+  const [open, setOpen] = useState(false);
```
oleksandr-danylchenko: `open` isn't descriptive enough in the context of the list component. "Who is opened? Is the list opened?"

frontend/src/components/administration/users/users-list.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const [selectedUser, setSelectedUser] = useState<User | undefined>();
+  const [updateUser] = useUpdateUserMutation();
+
+  const handleClickEdit = (user: User): void => {
+    setOpen(true);
+    setSelectedUser(user);
+  };
```
oleksandr-danylchenko: Why would we separately store the `open` and `selectedUser`? What does it mean to have the `selectedUser` but `open=false`?

oleksandr-danylchenko (changes requested):

frontend/src/components/administration/administration.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { UsersList } from './users/users-list';
+
+export const Administration = (): React.ReactElement => {
+  const { data, isLoading, isSuccess } = useGetUsersQuery();
+
+  return (
+    <Box sx={{ padding: '20px' }}>
+      <h1>Admin panel</h1>
+      {isLoading ? (
```
nikita-remeslov: Try to simplify this logic, very hard to understand what is going on here

shared/src/common/types/user/user.ts
```diff
@@ -0,0 +1,24 @@
+enum Role {
```
nikita-remeslov: I think for enum we should create a different file. Also, I prefer to specify with prefix an enum like this if we use it only for the User interface
Example.: UserRole

shared/src/common/types/user/user.ts
```diff
@@ -0,0 +1,24 @@
...
+  user,
+  admin,
+}
+
+enum Sex {
```
nikita-remeslov: The same here

nikita-remeslov (changes requested):

shared/src/common/types/user/user.ts
oleksandr-danylchenko: Great suggestion!

frontend/src/components/administration/administration.tsx
o-rudych: Good point, thanks.

frontend/src/components/administration/administration.tsx
o-rudych: Thanks, fixed it.

frontend/src/components/administration/users/dialog-edit-user.tsx
o-rudych: Agree, thanks.

frontend/src/components/administration/users/users-list.tsx
o-rudych: That's true, thanks. Renamed it to `openedUserEditing`.

frontend/src/components/administration/users/users-list.tsx
o-rudych: Makes sense. I removed `open` at all, left only `selectedUser` in state.

backend/src/routes/user/user.router.ts
o-rudych: Good point, thanks.

shared/src/common/types/user/user.ts
o-rudych: Good point, thanks! I've redefined the types.

shared/src/common/types/user/user.ts
o-rudych: Done

oleksandr-danylchenko (dismissed):
This looks good now. Only the conflicts need to be resolved and it'll be merged

o-rudych:
Conflicts are resolved

PR #48, opened 2022-08-18
Global toast message on comparison click

setar92:
https://gyazo.com/72d1060f168ef4c0c6ebd2f2223b5374
https://gyazo.com/07e2b03f9a14772da39950c380a6f9f1


setar92:
https://gyazo.com/72d1060f168ef4c0c6ebd2f2223b5374
https://gyazo.com/07e2b03f9a14772da39950c380a6f9f1

frontend/src/components/modal-compare/modal.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const closeModal = (): void => {
+    setIsHidden(!isHidden);
+  };
+
+  return (
+    <div className={styles.container} hidden={isHidden}>
+      <div className={styles.content}>
+        <button className={styles.close} onClick={(): void => closeModal()}>
+          ×
+        </button>
```
oleksandr-danylchenko: This `onClick` handler can be simplified. If you pass the anon func to the `onClick`, you can directly call the `setIsHidden`. But if you already have the `closeModal` defined, you can pass it directly to the `onClick`

```suggestion
        <button className={styles.close} onClick={closeModal}>
          ×
        </button>
```

frontend/src/components/modal-compare/modal.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <div className={styles.icon}>
+          <img
+            className={styles.button}
+            src={fillCompare}
+            alt="compare button"
+          />
```
oleksandr-danylchenko: **`img` cannot be the button**. And `div` cannot be a button. Only the `button` tag can be the button :slightly_smiling_face:. And the `button` can wrap the `img`

frontend/src/components/modal-compare/modal.tsx
```diff
@@ -0,0 +1,48 @@
...
+        </div>
+        <div className={styles.message}>
+          <p className={styles.inline}>{'You added '}</p>
+          <p className={styles.carName}>{carName}</p>
+          <p className={styles.inline}>{' to the Comparison!'} </p>
```
oleksandr-danylchenko: Please use `span` here, because it's already `inline`.

```suggestion
         You added{' '}
          <span className={styles.carName}>{carName}</span>{' '}
          to the Comparison!
```

frontend/src/components/modal-compare/styles.module.scss
```diff
@@ -0,0 +1,84 @@
...
+  display: inline-block;
+  color: var(--blue100);
+  text-transform: uppercase;
+}
+
+@media (max-width: 440px) {
+  .container {
+    right: 0;
+  }
+}
```
oleksandr-danylchenko: Please add some side margin for the card when it's used on mobile. It should stick to the sides of the screen

oleksandr-danylchenko (changes requested):
This looks good! I have only some codestyle improvements suggestions

setar92:
https://gyazo.com/b4e4958a444e5b2dd7f7b676d10ffb2d

frontend/src/components/modal-compare/modal.tsx
```diff
@@ -0,0 +1,50 @@
...
+import styles from './styles.module.scss';
+
+interface ModalProps {
+  carName: string;
+  carDescription: string;
+  isHidden: boolean;
+  setIsHidden: Dispatch<boolean>;
+}
+
+const Modal: FC<ModalProps> = ({
```
oleksandr-danylchenko: This component is the `TOAST`, not the `MODAL`.
Toast is a non-intrusive, optional message for the user. But the modal totally intrusive visual component which requires interaction from the user and blocks out the rest of the page

Toast example:
image

Modal: 
image

--- 

### So this component should be named `CompareModal`.

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko:
SORRY! I meant `CompareToast` :man_facepalming::man_facepalming::man_facepalming:

PR #49, opened 2022-08-18
Added edit profile backend

AlexanderKhapchenko:


backend/src/controllers/update-user/update-user.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+import { UpdateUserDto } from '@dtos/user/update-user.dto';
+import { User } from '@prisma/client';
+import * as updateUserService from '@services/update-user/update-user.service';
+import { NextFunction, Response } from 'express';
+import httpStatus from 'http-status-codes';
+
+export interface UpdateUser extends UserCreateInput {
+  tokenPayload: TokenPayload;
+  new_password: string;
+  repeat_new_password: string;
```
oleksandr-danylchenko: Please don't use `snake_case` variables

backend/prisma/schema.prisma
```diff
@@ -23,7 +24,7 @@ model User {
 
 model User_Security {
   id                     String  @id @default(uuid())
-  user                   User    @relation(fields: [user_id], references: [id])
+  user                   User    @relation(fields: [user_id], references: [id], onDelete: Cascade)
```
oleksandr-danylchenko, line 27: Makes sense! Thanks!

backend/src/controllers/update-user/update-user.controller.ts
oleksandr-danylchenko: Please use the `camelCase` for the variables definition

backend/src/controllers/update-user/update-user.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const updateUser = async (
+  req: TypedRequestBody<UpdateUser>,
+  res: Response<User>,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const user = UpdateUserDto.createFromRequest(req);
+
+    const { new_password, ...userData } = user;
```
oleksandr-danylchenko: `camelCase`

backend/src/middlewares/update-user/update-user.middleware.ts
```diff
@@ -0,0 +1,73 @@
...
+      throw new Error(ExceptionMessage.USER_EXISTS);
+    }
+  }
+};
+
+const passwordValidate = async (
+  req: TypedRequestBody<UpdateUser>,
+): Promise<void> => {
+  if (req.body.password) {
+    if (req.body.new_password != req.body.repeat_new_password) {
```
oleksandr-danylchenko: Don't use `==` or `!=` in modern JS. It's not safe because of the auto type coercion

backend/src/middlewares/update-user/update-user.middleware.ts
```diff
@@ -0,0 +1,73 @@
...
+    await passwordValidate(req);
+
+    next();
+  } catch (err) {
+    const { message } = err as ErrorMessage;
+    console.error(err);
+    res.status(httpStatus.FORBIDDEN).json({ error: message });
+    next(err);
+  }
+};
```
oleksandr-danylchenko: I wouldn't do the validation of the input values for the update in the middleware. It's better to be done in the service because the validation here can be a part of the business logic. Especially the prohibition to pass the already existing password

backend/src/services/password-validation/password-validation.service.ts
```diff
@@ -0,0 +1,11 @@
+import { User_Security } from '@prisma/client';
+import { compare as bcryptCompare } from 'bcrypt';
+
+const validatePassword = async (
+  password: string,
+  userSecurity: User_Security | null | undefined,
+): Promise<boolean> => {
+  return bcryptCompare(password, userSecurity?.password || '');
+};
```
oleksandr-danylchenko: Curly braces are redundant for the one-line funcs:
```suggestion
const validatePassword = async (
  password: string,
  userSecurity: User_Security | null | undefined,
): Promise<boolean> =>
  bcryptCompare(password, userSecurity?.password || '');
```

backend/src/services/update-user/update-user.service.ts
```diff
@@ -0,0 +1,58 @@
+import { prisma } from '@data/prisma-client';
+import { bcryptHash } from '@helpers/crypto/crypto';
+import { Prisma, User } from '@prisma/client';
+
+const updateUser = async (
+  user: Prisma.UserCreateInput,
+  new_password?: string,
+): Promise<User> => {
+  const { id, ...userData } = user;
+  console.log(userData);
```
oleksandr-danylchenko: Remove the `console.log`

backend/src/services/update-user/update-user.service.ts
```diff
@@ -0,0 +1,58 @@
+import { prisma } from '@data/prisma-client';
+import { bcryptHash } from '@helpers/crypto/crypto';
+import { Prisma, User } from '@prisma/client';
+
+const updateUser = async (
+  user: Prisma.UserCreateInput,
+  new_password?: string,
```
oleksandr-danylchenko: It would be better to pass the `new_password` as part of the `user` object. Because it's actually its part. Then, you can exclude it from the `userData` as const { id, password, ...userData } = user`

backend/src/services/update-user/update-user.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    },
+  });
+
+  return updatedUser;
+};
+
+const updatePassword = async (
+  newPassword: string,
+  id?: string,
+): Promise<void> => {
```
oleksandr-danylchenko: Generally, it's better to order arguments from less specific to more specific. So I would pass the `id` of the user as the first arg

oleksandr-danylchenko (changes requested):
In general, this PR looks good. Have some recommendations on the code style

backend/src/middlewares/update-user/update-user.middleware.ts
AlexanderKhapchenko: So we can't pass the current password in the reqest?
Then how do I know that the user has entered the current password correctly?

oleksandr-danylchenko: No, it's okay to pass it. But I just wouldn't check in the middleware

oleksandr-danylchenko (approved):

PR #61, opened 2022-08-22
Header wishlist integration

iryb:
Task https://app.asana.com/0/1202525679663009/1202720801120417

## One model in wishlist:
wishlist-header

## Empty wishlist
wishlist-0-header


oleksandr-danylchenko:
We mustn't show empty lists like that: 
image
It should be either emptiness or an explicit message like "Nothing is in wishlist". I would prefer to the latter

oleksandr-danylchenko:
Please choose the branch from which you started the current one as the merge target. In this way, the change diff will be much smaller

frontend/src/components/header/header.tsx
```diff
@@ -1,14 +1,18 @@
...
 import { Navigate } from './navigate/navigate';
 import { PrivateElements } from './private-elements/private-elements';
 import { PublicElements } from './public-elements/public-elements';
 
 export const Header: React.FC = () => {
+  const { data: wishlist } = useGetWishlistsQuery();
+
   const user = {
-    favorites: 3,
+    favorites: wishlist?.count as number | 0,
```
oleksandr-danylchenko: `|` here is not what you expect :sweat_smile: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_OR

oleksandr-danylchenko (changes requested):

iryb:
> We mustn't show empty lists like that: image It should be either emptiness or an explicit message like "Nothing is in wishlist". I would prefer to the latter

### If nothing is in favorites:
image

backend/src/services/preferences/wishlist.service.ts
```diff
@@ -253,9 +253,13 @@ const getWishlistByUserId = async (
...
   );
 
+  const wishlistCount =
+    sanitizedModelsWishlist.length + sanitizedComplectationsWishlist.length;
+
   const wishlists: WishlistsResponseDto = {
     models: sanitizedModelsWishlist,
     complectations: sanitizedComplectationsWishlist,
+    count: wishlistCount,
   };
```
oleksandr-danylchenko: I would put the counting login on the client side. Then we'll be able to update the liked count right after the like

frontend/src/components/header/header.tsx
```diff
@@ -1,14 +1,18 @@
...
 import { Navigate } from './navigate/navigate';
 import { PrivateElements } from './private-elements/private-elements';
 import { PublicElements } from './public-elements/public-elements';
 
 export const Header: React.FC = () => {
+  const { data: wishlist = 0 } = useGetWishlistsQuery();
+
   const user = {
-    favorites: 3,
+    favorites: wishlist && wishlist.count,
```
oleksandr-danylchenko: We shouldn't use `smth &&` check in the modern JS. There a optional [chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) in JS to work with possibly `undefined` objects

shared/src/common/types/preferences/wishlist-response-dto.type.ts
```diff
@@ -60,6 +60,7 @@ type ComplectationResponseDto = {
 interface WishlistsResponseDto {
   models?: ModelResponseDto[];
   complectations?: ComplectationResponseDto[];
+  count: number;
```
oleksandr-danylchenko: `count` would better live on the client

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #65, opened 2022-08-23
fix_frontent_added_missing_import

setar92:
just added one import.


oleksandr-danylchenko:
Please name PRs in plain English, preferably in Past Simple form. Smth like "Added missing import of `VerificationLinkPath`". In that way, it would be much clearer for the reviewer what was actually done

oleksandr-danylchenko (approved):

PR #66, opened 2022-08-23
Add Complete Set component

iryb:
Task https://app.asana.com/0/1202525679663009/1202838302178341

complete-set-table


iryb:
What do you think about white color box? I styled it as in mockup but the border is almost invisible on un-hover table row. Do you want to leave it as it is now or apply darker border?

oleksandr-danylchenko:
> What do you think about white color box?

It should be framed into the `box-shadow`, not the `border`. The result can look smth like this - https://playcode.io/946847/

frontend/src/common/enums/car/options.enum.ts
```diff
@@ -0,0 +1,12 @@
...
+  Complectation: 'Complectation',
+  Color: 'Color',
+  Motor: 'Motor',
+  WheelDrive: 'Wheel Drive',
+  Race: 'Race',
+  Price: 'Price',
+  Compare: 'Compare',
+};
+
+export { CarSetOptions };
```
oleksandr-danylchenko: This is not `enum`. Why it's located in the `enum` file? :eyes:

frontend/src/components/complete-set-table/complete-set-table.tsx
```diff
@@ -0,0 +1,98 @@
...
+                {car.brand} {car.model}
+              </TableCell>
+              <TableCell className={styles.tableRow}>
+                {car.complectation}
+              </TableCell>
+              <TableCell className={styles.tableRow}>
+                <div
+                  className={styles.colorBox}
+                  style={{ backgroundColor: car.color }}
+                ></div>
```
oleksandr-danylchenko: `div` can be collapsed: 
```suggestion
                <div
                  className={styles.colorBox}
                  style={{ backgroundColor: car.color }}
                />
```

oleksandr-danylchenko (changes requested):
- The colors boxes should be framed into the `box-shadow`
- We mustn't use objects instead of the `enum`s in TS

frontend/src/common/enums/car/options.enum.ts
iryb: You are totally right. Thank you!

oleksandr-danylchenko (approved):

PR #67, opened 2022-08-24
Added middleware for Roles management by admin

o-rudych:


backend/src/controllers/users/users.controller.ts
```diff
@@ -22,15 +33,16 @@ const getUsers = async (
...
   res: Response,
   next: NextFunction,
 ): Promise<void> => {
   try {
     const { id: userId } = req.params;
-    const userData = req.body;
+    const { tokenPayload, ...userData } = req.body;
     const user = await usersService.updateUser(userId, userData);
 
+    console.log(tokenPayload);
```
o-rudych: I think this is not the best way to get rid of the `tokenPayload` here and pass new object further. But for now, this is what I did. I will be grateful for advice.

oleksandr-danylchenko: No, it's a valid way to get stip off some properties from an object

oleksandr-danylchenko: This log should be removed I guess
```suggestion
```

oleksandr-danylchenko (dismissed):

oleksandr-danylchenko (approved):

PR #68, opened 2022-08-24
Edit profile frontend

AlexanderKhapchenko:


frontend/src/components/edit-profile/select-years-range/select-years-range.tsx
```diff
@@ -0,0 +1,52 @@
...
+  end: number;
+  required: boolean;
+  id?: string;
+  name: string;
+  control: Control;
+  label: string;
+  defaultValue: string;
+}
+
+export const SelectYearsRange: React.FC<RangeYearsProps> = ({
```
oleksandr-danylchenko: oh-oh-oh
The naming is totally misleading. Because as result we select the birthday **_year_**. A **_single_** unit of data, not the _**range**_. Also, as the component is named "...Years..." the years' range limitation can be an internal part of the component instead of using the props to pass the range

frontend/src/store/queries/user/update-user.ts
```diff
@@ -0,0 +1,44 @@
...
+  endpoints: (builder) => ({
+    updateUserProfile: builder.mutation<
+      ProfileFieldsResponseData,
+      ProfileFieldsRequestData
+    >({
+      query: ({ ...put }) => ({
+        url: `${API.USER}`,
+        method: 'PUT',
+        body: put,
+      }),
```
oleksandr-danylchenko: Why did you do the `{ ...put }`? 

const a = { foo: 'bar' };
const b = { ...a };

Both `a` and `b` have the same exact content

frontend/src/components/edit-profile/edit-profile.tsx
```diff
@@ -0,0 +1,257 @@
...
+  };
+
+  return (
+    <div className={styles.popupWrapper}>
+      <DialogDeleteAccount
+        handleCloseDialog={handleCloseDialog}
+        handleDeleteProfile={handleDeleteProfile}
+        isOpen={openDialog}
+      />
+      <div className={styles.popup}>
```
oleksandr-danylchenko, line 112: This is not the popup, it's the modal window. And for the `Modal` you could use the MUI's `Modal` component.

oleksandr-danylchenko (changes requested):

frontend/src/components/edit-profile/edit-profile.tsx
```diff
@@ -103,13 +103,19 @@ export const EditProfile: React.FC<EditProfileProps> = ({ onClose }) => {
   };
 
   return (
-    <div className={styles.popupWrapper}>
-      <DialogDeleteAccount
-        handleCloseDialog={handleCloseDialog}
-        handleDeleteProfile={handleDeleteProfile}
-        isOpen={openDialog}
-      />
+    // <div className={styles.popupWrapper}>
```
oleksandr-danylchenko: Remove the commented code

oleksandr-danylchenko (dismissed):

oleksandr-danylchenko:
Please resolve conflicts and it'll be merged

frontend/src/components/edit-profile/edit-profile.tsx
```diff
@@ -0,0 +1,262 @@
...
+        email: user.email,
+      },
+      validationSchema: updateUserSchema,
+    });
+
+  useEffect(() => {
+    if (deleteIsSuccess) {
+      navigate(AppRoute.ROOT);
+    }
+  });
```
oleksandr-danylchenko: You didn't use the `useEffect` dependencies. This effect will run on each rerender

frontend/src/components/edit-profile/dialog-delete-account/dialog-delete-account.tsx
```diff
@@ -0,0 +1,46 @@
...
+  DialogContent,
+  DialogContentText,
+  DialogTitle,
+} from '@mui/material';
+
+interface DialogDeleteAccountProps {
+  handleCloseDialog: () => void;
+  handleDeleteProfile: () => Promise<void>;
+  isOpen: boolean;
+}
```
oleksandr-danylchenko: The order of the props should go from the plain properties to the callbacks

oleksandr-danylchenko (changes requested):

frontend/src/common/types/button/button.ts
```diff
@@ -2,4 +2,5 @@ export type ButtonPropsType = {
   className?: string;
   text: string;
   onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  type?: 'button' | 'submit' | 'reset' | undefined;
```
oleksandr-danylchenko: Why would we need a `reset` type, it's not part of the HTML button types? And the `undefined` shouldn't be explicitly specified if you use the `?`. It's a duplication

frontend/src/components/edit-profile/select-field-form/select-field-form.type.ts
```diff
@@ -0,0 +1,18 @@
...
+  id?: string;
+  onChange?: (event: SelectChangeEvent) => void;
+  onClose?: (event: SelectChangeEvent) => void;
+  className?: string;
+  required: boolean;
+  errors?: string;
+  control: Control;
+  defaultValue?: string;
+  label: string;
+};
```
oleksandr-danylchenko: Please put callbacks in the end

oleksandr-danylchenko (changes requested):

shared/src/common/types/user/update-user.ts
```diff
@@ -0,0 +1,24 @@
...
+  email: string;
+}
+
+interface ProfileFieldsRequestData extends UserFields {
+  password: string;
+  newPassword: string;
+  repeatNewPassword: string;
+}
+
+interface ProfileFieldsResponseData extends UserFields {}
```
oleksandr-danylchenko: Empty interfaces that extend smth should `type`s
```suggestion
type ProfileFieldsResponseData = UserFields;
```

oleksandr-danylchenko (approved):
LGTM

PR #70, opened 2022-08-24
Added global style variables from Figma design

stanislav-ostr:


frontend/src/assets/css/typography.scss
```diff
@@ -1,22 +1,49 @@
...
+  font-weight: var(--font-weight-regular);
+  font-size: 0.875rem; // 14px
+}
+
+.tub {
+  font-weight: var(--font-weight-semibold);
+  font-size: 1.125rem; // 18px
+}
+
+.toggel {
```
oleksandr-danylchenko: `.toggle`?

oleksandr-danylchenko: Also check where this class is used

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko:
@stanislav-ostr noticed that the font size for the H3 (58px => 40px) and H4 (22px => 24px) don't match 
image

oleksandr-danylchenko (approved):

PR #75, opened 2022-08-24
Add filtered car list item

iryb:
Task https://app.asana.com/0/1202525679663009/1202838947527271

Note: That's how Car list item component looks with 14px font size inside 1200px container and 70% column. The rest 30% of space will be filled with Filters. See screenshot from Figma. As for me we have too much information in this card or too small container. Making fonts less then 14px may be harmfull for user experience.

filtered-car-list-item

### Figma
image


oleksandr-danylchenko:
@iryb 
Can we show just the engines' displacement here, not the actual names of the engines?
image

oleksandr-danylchenko:
To always show the scrollbar I would pass only 5 complications to the table when the `Collapse` isn't uncovered. In that way, there would be no overflow. But when the user clicks +3, all the complications could be passed

oleksandr-danylchenko (changes requested):
I would like to see the scrollbar before the `Collapse` is opened

iryb:
@oleksandr-danylchenko Thanks for your suggestion. The one thing I have concerns about is when Collapse button is clicked to hide the table rows, it jumps.

I also made Model column longer to prevent overflow by Collapse button (when model is displayed in two rows, bottom part is cut off).

chrome-capture-2022-7-25

oleksandr-danylchenko (approved):

PR #76, opened 2022-08-24
Add google auth

R0ixy:
Asana - https://app.asana.com/0/1202525679663009/1202829131337117/f

Implemented only functionality on sign in/up pages. 

Waiting for edit page to be ready to implement the rest. Most functions are already designed to work on edit page as well.
(Will be another PR)


backend/prisma/schema.prisma
```diff
@@ -26,14 +26,14 @@ model User_Security {
...
+  password               String? @db.VarChar(300)
   password_change_token  String? @db.VarChar(300)
   email_activation_token String? @unique @db.VarChar(300)
   email_change_token     String? @db.VarChar(300)
   email_provisional      String? @db.VarChar(300)
   refresh_token          String? @db.VarChar(300)
-  google_acc_id          String? @db.VarChar(300)
-  facebook_acc_id        String? @db.VarChar(300)
+  google_acc_id          String? @unique @db.VarChar(300)
+  facebook_acc_id        String? @unique @db.VarChar(300)
```
oleksandr-danylchenko, line 37: Makes sense :+1:

backend/src/middlewares/auth/strategies/google.strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+    try {
+      user = await prisma.user.findFirst({
+        where: {
+          User_Security: {
+            google_acc_id: profile.id,
+          },
+        },
+      });
+      if (!user) {
+        const payload =
```
oleksandr-danylchenko, line 26: Please use guard clauses instead of having redundant if-else nesting -https://learningactors.com/javascript-guard-clauses-how-you-can-refactor-conditional-logic/

oleksandr-danylchenko: Please use guard clauses instead of having redundant if-else nesting -https://learningactors.com/javascript-guard-clauses-how-you-can-refactor-conditional-logic/

backend/src/middlewares/auth/strategies/google.strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+    let user;
+    try {
+      user = await prisma.user.findFirst({
+        where: {
+          User_Security: {
+            google_acc_id: profile.id,
+          },
+        },
+      });
+      if (!user) {
```
oleksandr-danylchenko: Is this statement correct? Why do we try to update the `user` if it wasn't found?

frontend/src/components/sign/components/redirect-after-sign/redirect-after-sign.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const refreshToken = searchParams.get('refreshToken');
+
+  useEffect(() => {
+    if (accessToken && refreshToken) {
+      localStorage.setItem(StorageKey.TOKEN, accessToken);
+      localStorage.setItem(StorageKey.REFRESH, refreshToken);
+    }
+    window.close();
+  }, []);
+  return (
```
oleksandr-danylchenko: [Let it breathe](https://youtu.be/m7jfM9A6R-k?t=56) 
```suggestion
  }, []);

  return (
```

frontend/src/components/sign/components/sign-with-google/sign-with-google.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const onStorageEvent = (): void => {
+    const accessToken = localStorage.getItem('token');
+    dispatch(setCredentials({ accessToken }));
+    navigate('/', { replace: true });
+  };
+
+  const onGoogleSign = (): void => {
+    openGooglePage();
+    window.addEventListener('storage', onStorageEvent, false);
+  };
```
oleksandr-danylchenko: You must also remove the listener when the component will be unmounted. So I would prefer to register the event listener in the `useEffect` instead

frontend/src/components/sign/components/sign-with-google/sign-with-google.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const SignWithGoogle = ({ title }: { title: string }): React.ReactElement => {
+  const navigate = useNavigate();
+  const dispatch = useAppDispatch();
+
+  const onStorageEvent = (): void => {
+    const accessToken = localStorage.getItem('token');
+    dispatch(setCredentials({ accessToken }));
+    navigate('/', { replace: true });
+  };
```
oleksandr-danylchenko: Please react to the actual event here and check what is set into the LS. Because now this listener will _always_ fire on any changes in the LS

frontend/src/helpers/sign/open-google.ts
```diff
@@ -0,0 +1,11 @@
+import { ENV } from '@common/enums/app/env.enum';
+
+const openGooglePage = (token: string | null = null): void => {
+  window.open(
+    `${ENV.API_PATH}/auth/google/sign?token=${token}`,
+    'Google OAuth',
+    'height=615,width=605',
+  );
+};
```
oleksandr-danylchenko:
```suggestion
const openGoogleAuthPage = (token: string | null = null): void =>
  window.open(
    `${ENV.API_PATH}/auth/google/sign?token=${token}`,
    'Google OAuth',
    'height=615,width=605',
  );
```

oleksandr-danylchenko (changes requested):
Good job! But it requires some polishing before being merged in

backend/src/middlewares/auth/strategies/google.strategy.ts
R0ixy: Well, user wasn't found by Google ID, but if we recieved jwt it means that user exists and can be found by uuid so we can simply add Google ID to profile

backend/src/middlewares/auth/strategies/google.strategy.ts
```diff
@@ -0,0 +1,79 @@
...
+            User_Security: {
+              update: {
+                google_acc_id: profile.id,
+              },
+            },
+          },
+        });
+
+        return done(null, user);
+      }
```
oleksandr-danylchenko, line 47: Okay, the guard clauses really helped and now it's much clearer what happens if the user was found by JWT

frontend/src/components/sign/components/redirect-after-sign/redirect-after-sign.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const accessToken = searchParams.get('accessToken');
+  const refreshToken = searchParams.get('refreshToken');
+
+  useEffect(() => {
+    if (accessToken && refreshToken) {
+      localStorage.setItem(StorageKey.TOKEN, accessToken);
+      localStorage.setItem(StorageKey.REFRESH, refreshToken);
+    }
+    window.close();
+  }, []);
```
oleksandr-danylchenko: You didn't pass the `accessToken` & `refreshToken` as the dependencies to the `useEffect`. Therefore, this `useEffect` will run _only_ on the first render, where the `at` and `rt` are `undefined`

frontend/src/components/sign/components/sign-in-form/sign-in-form.tsx
```diff
@@ -43,7 +43,7 @@ export const SignInForm = (): React.ReactElement => {
         navigate(AppRoute.ROOT);
       });
   };
-  console.log(error);
+  if (error) console.log(error);
```
oleksandr-danylchenko: Please put it in `useEffect`

frontend/src/components/sign/components/sign-with-google/sign-with-google.tsx
```diff
@@ -0,0 +1,36 @@
...
+    openGoogleAuthPage();
+  };
+
+  useEffect(() => {
+    window.addEventListener('storage', onStorageEvent, false);
+
+    return (): void => {
+      window.removeEventListener('storage', onStorageEvent, false);
+    };
+  }, []);
```
oleksandr-danylchenko: **`removeEventListener` won't work** because on each render the `onStorageEvent` reference will be different. You need to wrap the `onStorageEvent` in `useCallback` and pass it as the dependency of the `useEffect`

oleksandr-danylchenko (changes requested):

frontend/src/components/sign/components/redirect-after-sign/redirect-after-sign.tsx
oleksandr-danylchenko: The question got resolved here - https://binary-academy-2022.slack.com/archives/D03TF60GL8H/p1661419055643939

frontend/src/components/sign/components/sign-in-form/sign-in-form.tsx
```diff
@@ -38,12 +39,16 @@ export const SignInForm = (): React.ReactElement => {
     signIn(user)
       .unwrap()
       .then(({ accessToken }: SignInResponseData) => {
-        localStorage.setItem('access-token', accessToken);
+        localStorage.setItem(StorageKey.TOKEN, accessToken);
```
oleksandr-danylchenko, line 42: :+1:

oleksandr-danylchenko (approved):

PR #81, opened 2022-08-25
Fix font sizes

iryb:
### Font sizes before fix:
font-sizes-before

### Font sizes fixed:
font-sizes-fixed

Also I added **Accout** nav link that leads to Personal page. Do we need **Settigns** link? Because user can edit his profile from the Personal page.
image

### DIFF
--- a/frontend/src/components/header/logo/logo.tsx
+++ b/frontend/src/components/header/logo/logo.tsx
@@ -1,13 +1,17 @@
 import React from 'react';
+import { Link } from 'react-router-dom';
 
 import LogoIcon from '@assets/images/logo.svg';
+import { AppRoute } from '@common/enums/enums';
 
 import styles from './logo.module.scss';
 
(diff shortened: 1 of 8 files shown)


frontend/src/components/header/logo/logo.tsx
```diff
@@ -1,13 +1,17 @@
...
 import styles from './logo.module.scss';
 
 export const Logo: React.FC = () => {
   return (
     <div className={styles.logoContainer}>
-      <img src={LogoIcon} alt="logo" />
+      <Link to={AppRoute.ROOT}>
+        <img src={LogoIcon} alt="logo" />
+      </Link>
     </div>
```
oleksandr-danylchenko, line 15: This change works, but it's not part of the font fix. Please keep PRs as atomized as possible

oleksandr-danylchenko (approved):

PR #83, opened 2022-08-25
Fix token initial state

iryb:
Issue: token was always null on inner pages, even it was in local storage.
Fix: I added check in initial state. If token is inside local storage we set initial state to use it.


frontend/src/store/auth/slice.ts
```diff
@@ -7,7 +7,7 @@ type AuthState = {
 
 const initialState: AuthState = {
   user: null,
-  token: null,
+  token: localStorage.getItem('access-token') || null,
```
oleksandr-danylchenko: Please use `StorageKey` added in - https://github.com/BinaryStudioAcademy/bsa-2022-autoline/pull/76

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #86, opened 2022-08-25
add characteristics list

stanislav-ostr:
example


stanislav-ostr:
**Also I had to fix `frontend/src/store/queries/cars 25:8` because slint didn't allow commit**

frontend/src/common/types/characteristics-list/characteristics-list.ts
```diff
@@ -0,0 +1,16 @@
...
+export type CharacteristicsGroupProps = {
+  name: string;
+  options: CharacteristicsOption[];
+};
+
+export type CharacteristicsOption = {
+  value: string;
+  name?: string;
+  color?: string;
+};
```
oleksandr-danylchenko: I have a few questions for this type:
1. What will each field contain if they all are defined?
2. Why do we have a separate field for the color? Isn't it just one of the options

stanislav-ostr: example2

stanislav-ostr: Would I should add one component with its own type (value, name, color) for the first list and a second component with the type (only value) for other lists?

oleksandr-danylchenko: Okay, after the screenshot it started to make sense. Could you please attack a link to it in the comment for the `CharacteristicsOption`?

oleksandr-danylchenko (approved):

PR #92, opened 2022-08-25
add cities endpoint

R0ixy:


backend/src/controllers/cars/cars.controller.ts
```diff
@@ -66,4 +66,23 @@ const getUsedOptions = async (
...
 };
 
-export { carsSearchAutoria, getBrands, getModelsOfBrand, getUsedOptions };
+const getCitiesOfRegion = async (
+  req: TypedRequestParams<{ id: string }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const models = await carsService.getCities(req.params.id);
```
oleksandr-danylchenko: Cities are not part of the `car`'s ecosystem. They should be handled in a separate service

backend/src/services/cars/cars.service.ts
```diff
@@ -62,4 +63,14 @@ const getUsedOptions = async (): Promise<UsedOptionsType> => {
...
-export { getBrands, getModels, getUsedOptions };
+const getCities = async (regionId: string): Promise<Partial<City>[]> => {
+  return await prisma.city.findMany({
+    where: { region_id: regionId },
+    select: {
+      id: true,
+      name: true,
+    },
+  });
+};
```
oleksandr-danylchenko: **Don't use return await in async functions!**

https://kanby.medium.com/%D0%BF%D0%BE%D1%87%D0%B5%D0%BC%D1%83-%D0%B8%D1%81%D0%BF%D0%BE%D0%BB%D1%8C%D0%B7%D0%BE%D0%B2%D0%B0%D1%82%D1%8C-return-await-%D0%BF%D0%BB%D0%BE%D1%85%D0%B0%D1%8F-%D0%B8%D0%B4%D0%B5%D1%8F-e87b70015f0c

oleksandr-danylchenko (changes requested):

backend/src/controllers/cars/cars.controller.ts
R0ixy: Well, but we handle regions in the cars service. This functionality was implemented eartier, I'm just adding some missing parts

oleksandr-danylchenko: Oh-oh. Then it somehow slipped through my attention. Could you please create a separate service for the locations?

oleksandr-danylchenko (approved):

PR #93, opened 2022-08-25
Create cars details panel

brovarenko:
See - https://app.asana.com/0/1202525679663009/1202804889720781/f
Complectation
complectation
Model
model


oleksandr-danylchenko:
Some screenshots would be appreciated

backend/src/controllers/complectations/complectations.controller.ts
```diff
@@ -0,0 +1,29 @@
...
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const { complectationId } = req.query;
+    const ResponseDto = await getComplectationsById(complectationId);
+    res.json(ResponseDto).status(httpStatus.CREATED);
+  } catch (error) {
+    console.error(error);
+    if (error instanceof Error) {
+      res.sendStatus(404);
```
RomanMahola: httpStatus.NOT_FOUND

error instanceof Error 
What if we will have "connection refused to database?"  We will also get 404 but it is not true

backend/src/controllers/complectations/complectations.controller.ts
```diff
@@ -0,0 +1,29 @@
...
+  req: TypedRequestQuery<ModelIdInput>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const { complectationId } = req.query;
+    const ResponseDto = await getComplectationsById(complectationId);
+    res.json(ResponseDto).status(httpStatus.CREATED);
+  } catch (error) {
+    console.error(error);
```
RomanMahola: Delete console.log

backend/src/services/complectations/complectations.service.ts
```diff
@@ -0,0 +1,85 @@
...
+    upholstery: [],
+    sound: [],
+    design: [],
+    comfort: [],
+    auxiliary: [],
+    important: [],
+  };
+
+  optionType.forEach((optionName) => {
+    data?.options.forEach((option) => {
```
RomanMahola: I prefer to use reduce here
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

RomanMahola (changes requested):

backend/src/controllers/complectations/complectations.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+  try {
+    const userId = req.body.tokenPayload.sub;
+    const { complectationId, modelId } = req.query;
+    const input: ComplectationsInput = {
+      userId,
+      modelId,
+      complectationId,
+    };
+    const ResponseDto = await getComplectationsById(input);
+    res.json(ResponseDto).status(httpStatus.OK);
```
oleksandr-danylchenko: Please use `camelCase` for vars

backend/src/controllers/complectations/complectations.controller.ts
```diff
@@ -0,0 +1,36 @@
+import { getComplectationsById } from '@services/complectations/complectations.service';
```
oleksandr-danylchenko: Please use wildcard import `import * as complectationsService from ...`

backend/src/services/complectations/complectations.service.ts
```diff
@@ -0,0 +1,225 @@
...
+        },
+        select: {
+          id: true,
+          complectation: {
+            select: {
+              id: true,
+            },
+          },
+        },
+      },
```
oleksandr-danylchenko: How wishlist impact getting of the complectation by its id? If we expect to return complectations only from the wishlist it should be expressed in the router and the name of the method

backend/src/services/complectations/complectations.service.ts
```diff
@@ -0,0 +1,225 @@
...
+      fuel_type: { select: { name: true } },
+      transmission_type: { select: { name: true } },
+      options: {
+        select: {
+          option: { select: { name: true, type: true } },
+          important: true,
+        },
+      },
+    },
+  })) as ComplectationReturnedDbType;
```
oleksandr-danylchenko: Why do we need explicit type coercion? Prisma already should generate a proper type

backend/src/services/complectations/complectations.service.ts
```diff
@@ -0,0 +1,225 @@
...
+    important: [],
+  };
+
+  const importantOptionsList: Set<string> = new Set();
+  data?.options.forEach((option) => {
+    if (option.important === true) {
+      importantOptionsList.add(option.option.name);
+    }
+  }),
+    (options.important = Array.from(importantOptionsList));
```
oleksandr-danylchenko, line 66: I don't understand what's going on here. You can also get only unique values using the `.filter()`

backend/src/services/complectations/complectations.service.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  const optionsList = data?.options.reduce((acc: OptionType, obj) => {
+    const key = obj.option['type'];
+    acc[key] ??= [];
+    acc[key].push(obj.option.name);
+    return acc;
+  }, {});
```
oleksandr-danylchenko: Please more explicit name for the accumulator

backend/src/services/complectations/complectations.service.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  const model = (await prisma.model.findUnique({
+    where: {
+      id: modelId,
+    },
+    select: {
+      users_wishlists: {
+        where: {
+          user_id: userId,
+        },
```
oleksandr-danylchenko, line 109: Why wishlist?

backend/src/services/complectations/complectations.service.ts
```diff
@@ -0,0 +1,225 @@
...
+        }
+        if (option.important === true) {
+          importantOptionsList.add(option.option.name);
+        }
+      }),
+    );
+    options.important = Array.from(importantOptionsList);
+    switch (optionName) {
+      case 'security':
+        options.security = Array.from(optionsList);
```
oleksandr-danylchenko: You can use the newer spread operator `[...smth]`

backend/src/services/complectations/complectations.service.ts
```diff
@@ -0,0 +1,225 @@
...
+    data: model,
+    enginePowers: Array.from(enginePowers),
+    colors: Array.from(colors),
+    engineDisplacements: Array.from(engineDisplacements),
+    drivetrains: Array.from(drivetrains),
+    fuelTypes: Array.from(fuelTypes),
+    transmissionTypes: Array.from(transmissionTypes),
+    options: options,
+    priceStart: model?.prices_ranges[0].price_start,
+    priceEnd: model?.prices_ranges[0].price_end,
```
oleksandr-danylchenko: For the model's `priceEnd` you should use the `prices_ranges` of the last and the most expensive complectation

frontend/src/components/details-car-panel/details-car-panel.tsx
```diff
@@ -0,0 +1,195 @@
...
+    complectationId,
+    modelId,
+  });
+
+  const [createWishlist] = useCreateWishlistMutation();
+  const [deleteWishlist] = useDeleteWishlistMutation();
+  const dispatch = useAppDispatch();
+  useEffect(() => {
+    dispatch(fetchRates());
+  }, []);
```
oleksandr-danylchenko: Why didn't use the RTKQ hook instead of the custom async thunk?

frontend/src/components/details-car-panel/details-car-panel.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const rate = useAppSelector((state) => state.rates.rateUSD);
+  const handleCreateWishlist = async (): Promise<void> => {
+    const data: WishlistInput =
+      complectationId === '' ? { modelId } : { complectationId };
+
+    await createWishlist(data);
+  };
```
oleksandr-danylchenko: Please use empty lines and let the component breathe

frontend/src/store/exchange-rates/slice.ts
```diff
@@ -0,0 +1,35 @@
...
+  sale: string;
+};
+const initialState: ExchngeRateState = {
+  rateUSD: '',
+};
+
+export const fetchRates = createAsyncThunk('rates/fetchRates', async () => {
+  const response = await fetch(
+    'https://api.privatbank.ua/p24api/pubinfo?json&exchange&coursid=5',
+  );
```
oleksandr-danylchenko: I would rather put smwr at the root level as the constant value

frontend/src/store/exchange-rates/slice.ts
```diff
@@ -0,0 +1,35 @@
+import { createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+
+type ExchngeRateState = {
+  rateUSD: string;
+};
```
oleksandr-danylchenko: Use `interface` here

frontend/src/store/exchange-rates/slice.ts
```diff
@@ -0,0 +1,35 @@
...
+
+type ResponseThunkType = {
+  ccy: string;
+  base_ccy: string;
+  buy: string;
+  sale: string;
+};
```
oleksandr-danylchenko: Interface

frontend/src/store/exchange-rates/slice.ts
```diff
@@ -0,0 +1,35 @@
...
+const { reducer } = createSlice({
+  name: 'rates',
+  initialState,
+  reducers: {},
+  extraReducers: (builder) => {
+    builder.addCase(fetchRates.fulfilled, (state, action) => {
+      state.rateUSD = action.payload.filter((e) => e.ccy == 'USD')[0].sale;
+    });
+  },
+});
```
oleksandr-danylchenko: Why do we even need a slice? Why didn't use the RTKQ cache

shared/src/common/types/complectations/complectations.ts
```diff
@@ -0,0 +1,114 @@
...
+  };
+  options: {
+    important: boolean;
+    option: {
+      name: string;
+      type: string;
+    };
+  }[];
+} | null;
+
```
oleksandr-danylchenko: To use it on frontend we need to have a DTO type

oleksandr-danylchenko (changes requested):

backend/src/services/complectations/complectations.service.ts
brovarenko: We use this information to find out if this complectation is in the user's wish list in order to correctly display the like icon.

oleksandr-danylchenko:
@brovarenko 
Please fix the linter

oleksandr-danylchenko:
Okay, the changes make sense, but the options pills need to be reworked to not bloat the card. I'll bring it up on the daily call

backend/src/services/complectations/complectations.service.ts
```diff
@@ -0,0 +1,232 @@
...
+        break;
+      case 'optics':
+        options.optics = [...optionsList];
+        break;
+      case 'multimedia':
+        options.multimedia = [...optionsList];
+        break;
+      case 'upholstery':
+        options.upholstery = [...optionsList];
+        break;
```
oleksandr-danylchenko, line 176: Why do we need the [....smth]? Why do we need to have a copy of the array?

frontend/src/components/details-car-panel/details-car-panel.tsx
```diff
@@ -0,0 +1,194 @@
...
+    complectationId,
+    modelId,
+  });
+  const { data: rate } = useGetRateQuery();
+  const [createWishlist] = useCreateWishlistMutation();
+  const [deleteWishlist] = useDeleteWishlistMutation();
+
+  const handleCreateWishlist = async (): Promise<void> => {
+    const data: WishlistInput =
+      complectationId === '' ? { modelId } : { complectationId };
```
oleksandr-danylchenko: `''` is already considered nullish value, so it can be simplified to the `complectationId ? { complectationId } : { modelId }`

oleksandr-danylchenko (approved):

PR #94, opened 2022-08-25
Fix/changes in filter form

elador13:
[Removed Race, added engine power & displacement](https://github.com/BinaryStudioAcademy/bsa-2022-autoline/commit/3294ede3d4c396efa2431bb67b63447ceb45dc82)

[Fixed range-selector's local state behaviour after reset all filters](https://github.com/BinaryStudioAcademy/bsa-2022-autoline/commit/e1977a9d3dbd164dc3f862bb31736079b63d61db)


frontend/src/components/common/range-selector/range-selector.tsx
```diff
@@ -29,7 +29,12 @@ const RangeSelector: FC<Props> = (props) => {
   };
 
   useEffect(() => {
-    if (+maxValue && +minValue > +maxValue) {
+    setMinValue(props.selectedMin);
+    setMaxValue(props.selectedMax);
+  }, [props.selectedMin, props.selectedMax]);
+
+  useEffect(() => {
+    if (+minValue && +maxValue && +minValue > +maxValue) {
```
RomanMahola, line 37: Think how to simplify it

RomanMahola (changes requested):

frontend/src/components/common/range-selector/range-selector.tsx
elador13: I don't know how to simplify it. This check needs to swap values if min value higher tnan max, so we need to know if both values exists. For example minYear=2015, maxValue=2000 and in the next line they are swaped.

oleksandr-danylchenko: @elador13 
I think that we can convert props to numbers when we assign 'em to the `useState` and in `useEffect`

oleksandr-danylchenko (approved):

PR #112, opened 2022-08-28
Add advanced filter search button + some changes in simple form button

elador13:
See - https://app.asana.com/0/1202525679663009/1202867876332550/f

Screenshot_1
Screenshot_2
adv-filter-button


frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
```diff
@@ -129,7 +136,7 @@ const AdvancedAutoFilter: FC = () => {
               label="Regions"
               onChange={handleRegionChange}
               value={getValueById(options.regions, filters.regionId)}
-              options={options?.regions?.map((item: AutoRiaOption) => ({
+              options={options.regions.map((item: AutoRiaOption) => ({
```
oleksandr-danylchenko, line 149: Is the removal of the optional chaining expected? Do we know that the `options` and `regions` will be defined? Why the `?.` was added in the first place?

frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
```diff
@@ -236,6 +243,19 @@ const AdvancedAutoFilter: FC = () => {
...
+            <Button
+              onClick={doSearch}
+              className={styles.searchButton}
+              variant="contained"
+            >
+              SEARCH
+            </Button>
+          </div>
+        </Zoom>
+      )}
```
oleksandr-danylchenko, line 258: `isButtonVisible &&` will make the `Zoom` immediately unmounted and the out animation won't work

frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
```diff
@@ -236,6 +243,19 @@ const AdvancedAutoFilter: FC = () => {
         <UTurnRightIcon className={styles.resetIcon} />
         Reset All Filters
       </h6>
+      {isButtonVisible && (
+        <Zoom in={isButtonVisible}>
```
oleksandr-danylchenko: `isButtonVisible &&` will make the `Zoom` immediately unmounted and the out animation won't work

oleksandr-danylchenko (changes requested):

frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
elador13: I already use main check of existing options in line 126 `if (isLoading) return <h1>Loading...</h1>;`
Also I use conditional rendering in every mounting of those components.  Probably I added `?.`  before just to quickly fix TSerrors.

oleksandr-danylchenko (approved):

PR #113, opened 2022-08-28
fix tabs in drawer, new car card's styles

ashonuyak:
Used `/* stylelint-disable-next-line value-no-vendor-prefix */` to ignore this rule

### DIFF
--- a/frontend/src/components/new-car-card/styles.module.scss
+++ b/frontend/src/components/new-car-card/styles.module.scss
@@ -81,8 +81,11 @@
 }
 
 .carDescription {
-  max-height: 5.3rem;
+  /* stylelint-disable-next-line value-no-vendor-prefix */
+  display: -webkit-box;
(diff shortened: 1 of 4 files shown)


frontend/src/components/new-car-card/styles.module.scss
```diff
@@ -81,7 +81,9 @@
 }
 
 .carDescription {
+  max-height: 5.3rem;
   margin: 0;
+  overflow: hidden;
```
oleksandr-danylchenko: Why should we just hide the overflown text? Because it could be confusing. But instead, we should use should show `...` as there's some more text available. There's a CSS prop `line-clamp` which makes it pretty easy

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #116, opened 2022-08-28
Add fetching of the current user info

AlexanderKhapchenko:
See Asana Task:
https://app.asana.com/0/1202525679663009/1202867907163259/f


AlexanderKhapchenko:
I don't quite understand why there are so many comites shining here. If anything, you only need to look at the ones from today
Such as Added backend functionality, Fixed validation schema, Added query for user information etc.

backend/src/controllers/update-user/update-user.controller.ts
```diff
@@ -47,4 +48,21 @@ const deleteUser = async (
...
 };
 
-export { updateUser, deleteUser };
+const getUser = async (
+  req: TypedRequestBody<{ tokenPayload: TokenPayload }>,
+  res: Response<Partial<UpdateUserReq>>,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const user = await updateUserService.getUser(req.body.tokenPayload.sub);
```
oleksandr-danylchenko: Please rename `updateUserService` into more generic `userService`

backend/src/services/update-user/update-user.service.ts
```diff
@@ -82,6 +82,26 @@ const deleteUser = async (id: string): Promise<Partial<UpdateUserReq>> => {
...
+    ? {
+        name: user.name,
+        birthYear: user.birth_year,
+        sex: user.sex,
+        phone: user.phone,
+        email: user.email,
+        location: user.location,
+        photoUrl: user.photo_url,
+      }
+    : null;
```
oleksandr-danylchenko: I would rather throw an exception here instead of return ambiguous `null` value

backend/src/services/update-user/update-user.service.ts
```diff
@@ -82,6 +82,26 @@ const deleteUser = async (id: string): Promise<Partial<UpdateUserReq>> => {
   };
 };
 
+const getUser = async (id: string): Promise<Partial<UpdateUserReq> | null> => {
```
oleksandr-danylchenko, line 85: Why don't use the existing `UpdateUserDto` here?

frontend/src/components/header/default-elements/default-elements.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <Link to={AppRoute.SIGN_IN}>
+        <ButtonOutline className={styles.btnOutline} text="Sign In" />
+      </Link>
+    </>
+  ) : (
+    <Link to={AppRoute.SIGN_IN} className={styles.signupBtn}>
+      <img src={UserIcon} className={styles.authAvatar} />
+    </Link>
+  );
+};
```
oleksandr-danylchenko: I don't think that the user image is an intuitive identifier that the user can sign in/up. We should explicitly show Create account/Sign in buttons or user burger menu

frontend/src/components/header/default-elements/default-elements.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Link } from 'react-router-dom';
+
+import UserIcon from '@assets/images/header/default-avatar.png';
+import { AppRoute } from '@common/enums/app/app-route.enum';
+import { ButtonFill } from '@components/common/button-fill/button-fill';
+import { ButtonOutline } from '@components/common/button-outline/button-outline';
+import styles from '@components/header/styles.module.scss';
+import { useMediaQuery, useTheme } from '@mui/material';
+
+const DefaultElements: React.FC = () => {
```
oleksandr-danylchenko: Please use a more explicit name. Because `DefaultElement` doesn't have any semantic meaning or context

frontend/src/components/header/header.tsx
```diff
@@ -17,29 +32,54 @@ export const Header: React.FC = () => {
...
+          background: '#ffffff',
+          boxShadow: 0,
+        }}
+      >
+        <Toolbar>
+          <Link to={AppRoute.ROOT}>
+            <img className={styles.logo} src={Logo} alt="Autoline" />
+          </Link>
+          {isMatchSm ? (
+            <DrawerComp />
```
oleksandr-danylchenko: What is the `Comp`? Please be more explicit. If it means the `component` than we should restate already known fact

frontend/src/components/header/header.tsx
```diff
@@ -17,29 +32,54 @@ export const Header: React.FC = () => {
...
+              <Tabs
+                onChange={(e, value): void => setValue(value)}
+                value={value}
+                className={styles.nav}
+              >
+                <Tab label="Used Cars" className={styles.navLink} />
+                <Tab label="New Cars" className={styles.navLink} />
+                <Tab label="About us" className={styles.navLink} />
+              </Tabs>
+              {store.getState().auth.token ? (
```
oleksandr-danylchenko: Please use `useAppSelector` and avoid usage of the `store.getState()`. It'll make the component rerender on each change of the store

oleksandr-danylchenko (changes requested):

backend/src/services/update-user/update-user.service.ts
AlexanderKhapchenko: Because UpdateUserDto has an additional field newPassword

frontend/src/components/header/default-elements/default-elements.tsx
AlexanderKhapchenko: We had two headers in the project and I combined them into one
I think it's a component, so I renamed it now)

frontend/src/components/header/header.tsx
AlexanderKhapchenko: We had two headers in the project and I combined them into one
I think it's a component, so I renamed it now)

frontend/src/components/header/default-elements/default-elements.tsx
AlexanderKhapchenko: Yes, I combined two existing headers into one
I didn't do this adaptive.
I'll take the next task to make an adaptive hedgerow, as the authorised user's adaptive is bad

oleksandr-danylchenko (dismissed):

oleksandr-danylchenko:
@o-rudych 
I remember you faced the same CI issue. Could you help @AlexanderKhapchenko to proceed?

oleksandr-danylchenko (dismissed):

backend/api-autoria/cars/brands-with-models.ts
```diff
@@ -6,6 +6,7 @@ export const brands = [
     'eng': 'mazda',
     'marka_id': 47,
     'name': 'Mazda',
+    'slang': 'Мазда',
```
RomanMahola: ?

backend/api-autoria/cars/cars-data.ts
```diff
@@ -6,6 +6,7 @@ export const carsData = [
     'eng': 'mazda',
     'marka_id': 47,
     'name': 'Mazda',
+    'slang': 'Мазда',
```
RomanMahola: ?

frontend/src/components/header/drawer/styles.module.scss
```diff
@@ -0,0 +1,17 @@
...
+
+  span {
+    display: flex;
+    justify-content: center;
+  }
+
+  .logout {
+    position: absolute;
+    bottom: 0;
+
```
RomanMahola, line 12: logout 100%

RomanMahola (approved):

PR #120, opened 2022-08-28
added search by name for user management and fixed sending email after signup

o-rudych:


o-rudych:
@oleksandr-danylchenko, @RomanMahola, could you please review my PR?

backend/src/services/users/users.service.ts
```diff
@@ -1,8 +1,21 @@
...
+        name: {
+          contains: username,
+        },
+      },
+    });
+  } else {
+    data = await prisma.user.findMany();
+  }
+
+  return data;
```
oleksandr-danylchenko: You don't need to create a variable here. Especially, with such a generic name that doesn't convey any context.

You can just use return before `findMany`

frontend/src/components/administration/administration.tsx
```diff
@@ -1,26 +1,52 @@
...
+      setDebouncedValue(value);
+    }, delay);
+
+    return () => {
+      clearTimeout(handler);
+    };
+  }, [value, delay]);
+
+  return debouncedValue;
+};
```
oleksandr-danylchenko: This must be outside of the component file, inside some shared hooks directory preferably. Because this hook can be reused in multiple places

frontend/src/components/administration/users/dialog-edit-user.tsx
```diff
@@ -66,80 +66,82 @@ const DialogEditUser: FC<DialogEditUserProps> = (props) => {
...
+          <Container>
+            {role && (
+              <FormControl sx={{ width: 300 }}>
+                <InputField
+                  name="name"
+                  type="text"
+                  inputLabel="Full name"
+                  value={name}
+                  onChange={handleChangeName}
+                />
```
oleksandr-danylchenko, line 82: I've just noticed a big mistake in this component. You don't use the react-hook-form, but `useState`. Please look at how the form and validation done for the sign-in and migrate that approach here

oleksandr-danylchenko (changes requested):

backend/src/services/users/users.service.ts
o-rudych: I've read before that this is bad practice to use `return await`. That's why I did this.
@oleksandr-danylchenko, what do you think about?
https://kanby.medium.com/%D0%BF%D0%BE%D1%87%D0%B5%D0%BC%D1%83-%D0%B8%D1%81%D0%BF%D0%BE%D0%BB%D1%8C%D0%B7%D0%BE%D0%B2%D0%B0%D1%82%D1%8C-return-await-%D0%BF%D0%BB%D0%BE%D1%85%D0%B0%D1%8F-%D0%B8%D0%B4%D0%B5%D1%8F-e87b70015f0c

oleksandr-danylchenko: That's right. We mustn't use `return await`. But, you should just return a promise. 
`return prisma.user...` is the expected return format. Also, it'll allow avoiding redundant if-else nesting

o-rudych: I got it. Fixed.

frontend/src/components/administration/administration.tsx
o-rudych: Yes, sure. That will be more correct. Done.

RomanMahola (approved):

PR #123, opened 2022-08-29
Create mobile layout for landing page

stanislav-ostr:
demo


oleksandr-danylchenko:
image
It's not 100% intuitive that the drawer opens on the left side when we click on the burger which is on the right side. Also, the right side of the screen is much easier to reach with the thumb.
Could you please make it open from the right side?

oleksandr-danylchenko (changes requested):
The code looks good :rocket: 
But could you only change the location of the drawer?

stanislav-ostr:
demo

o-rudych (approved):

PR #124, opened 2022-08-29
Fix general content width

iryb:
Fixed general content width on all pages except Landing page.

## Before:
content-width-before

## After: 
content-width-after


oleksandr-danylchenko (approved):
Thanks :+1:

PR #144, opened 2022-08-30
Fix/Bug_18 Search results empty when no filter selected

elador13:
search2


frontend/src/helpers/car-filters/is-filters-empty.ts
```diff
@@ -0,0 +1,7 @@
+export const isFiltersEmpty = (filters: {
+  [p: string]: string | string[];
+}): boolean => {
+  return Boolean(
+    Object.values(filters).every((filter) => filter === '' || !filter.length),
+  );
+};
```
oleksandr-danylchenko: [`.every()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) already returns a Boolean. And the `{}` are redundant

frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
```diff
@@ -61,6 +62,10 @@ const AdvancedAutoFilter: FC<AdvancedAutoFilterProps> = (props) => {
         modelId: brandDetails.map((item) => item.modelId),
       }),
     );
+
+    if (isFiltersEmpty(filters)) {
+      search([], true);
+    }
```
oleksandr-danylchenko: You're trying to reference the `search` method which is initialized beneath

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #149, opened 2022-08-30
Fixed bug 22: access to routes

iryb:
Task https://app.asana.com/0/1202525679663009/1202891560024280


frontend/src/navigation/routing/routing.tsx
```diff
@@ -12,23 +12,23 @@ import { ResetPassword } from '@components/reset-password/reset-password';
...
 
 const Routing: FC = () => {
-  const authData = {
-    name: 'Oleksandr',
-    role: 'admin',
-  };
-  const { role } = authData;
-  const isAdmin = role === 'admin';
+  const { data: authData } = useGetUserQuery();
+  const isAdmin = authData?.role === 'admin';
```
oleksandr-danylchenko: I'm afraid that this won't work as expected in the scenario when the admin page is reloaded or visited directly by the link :sweat_smile:. Because on the first render the `authData` is undefined and the `isAdmin` will never be `true`. Therefore, the `<Route element={<ProtectedRoute isAllowed={isAdmin} />}>` will redirect a probable admin to the sign-in page.

oleksandr-danylchenko (changes requested):

frontend/src/navigation/routing/routing.tsx
iryb: @oleksandr-danylchenko Yes, you are absolutely right. I would appreciate some ideas about other way to send user role request at router level. Right now I don't see any other options if we can't use local storage and redux state

R0ixy: I think we can use `jwt-decode` package to decode jwt on client so this way we will know role on the first render. The only problem is that we can't check if token is valid. But if it's not we will get an error while fetching and will be redireted from protected page.

Just a suggestion.

oleksandr-danylchenko: @R0ixy @iryb 
We mustn't get the role of the the user from token stored in the local storage. Because that token can be easily modified on the client side. Also, if the role was changed from the admin panel, the token won't know about it

oleksandr-danylchenko: Why won't we just wait till the user will be loaded and only then check if they are an admin? :eyes: 
And before the user will be loaded we can show a global loaded

oleksandr-danylchenko (changes requested):
https://github.com/BinaryStudioAcademy/bsa-2022-autoline/pull/149#discussion_r959348270

oleksandr-danylchenko (approved):

iryb:
@oleksandr-danylchenko these changes are not in `develop` branch even though `user-infomation` was merged into `develop`. I think I should push this code again in new PR.

oleksandr-danylchenko:
:eye: :nose: :eye: 
Whoa, sure! Please feel free to create a new PR :pray: . I believe that the `use-information` was dismissed because it was extremely large and we couldn't handle the errors.

PR #151, opened 2022-08-30
Add access token renewal

R0ixy:
Asana - https://app.asana.com/0/1202525679663009/1202879226111242/f


frontend/src/store/queries/index.ts
```diff
@@ -25,8 +28,49 @@ const baseQuery = fetchBaseQuery({
...
+  await mutex.waitForUnlock();
+  let result = await baseQuery(args, api, extraOptions);
+
+  if (result.error && result.error.status === 401) {
+    if (!mutex.isLocked()) {
+      const release = await mutex.acquire();
+      try {
+        const refreshResult = await baseQuery(
+          {
+            url: '/auth/refreshToken',
```
oleksandr-danylchenko: Please use API object

oleksandr-danylchenko (changes requested):

frontend/src/store/queries/index.ts
```diff
@@ -25,8 +28,49 @@ const baseQuery = fetchBaseQuery({
...
+  await mutex.waitForUnlock();
+  let result = await baseQuery(args, api, extraOptions);
+
+  if (result.error && result.error.status === 401) {
+    if (!mutex.isLocked()) {
+      const release = await mutex.acquire();
+      try {
+        const refreshResult = await baseQuery(
+          {
+            url: `${ApiPath.AUTH}/refreshToken`,
```
oleksandr-danylchenko: `/refreshToken` also should be part of the `ApiPath`

oleksandr-danylchenko (changes requested):

frontend/src/store/queries/index.ts
R0ixy: ouch, ok

frontend/src/common/enums/api/api-path.enum.ts
```diff
@@ -1,5 +1,6 @@
 enum ApiPath {
   AUTH = '/auth',
+  REFRESH_TOKEN = '/refreshToken',
```
oleksandr-danylchenko: These paths should include the full URL, so the `REFRESH_TOKEN` should be:
```suggestion
  AUTH = '/auth',
  REFRESH_TOKEN = '/auth/refreshToken',
```

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #156, opened 2022-08-31
Fixed Bug_28 - can't unlike the model when the page is not refreshed

iryb:
Task https://app.asana.com/0/1202525679663009/1202897278017233

Added retrieving the id of newly created wishlist. In this case we always have the actual wishlist id, no matter if page was refreshed or no.


oleksandr-danylchenko (changes requested):
I think that we need to have the delete request for the wishlist assembled a bit differently. It should receive the same parameters as the `createWishlist`. So there won't be a need to store a redundant `wishlistId`

oleksandr-danylchenko (approved):

PR #158, opened 2022-08-31
Add simple filter form in the Home Page when user authorized

brovarenko:
See - https://app.asana.com/0/1202525679663009/1202894582381808/f


frontend/src/components/landing-page/landing-page.tsx
```diff
@@ -37,19 +39,24 @@ export const LandingPage = (): React.ReactElement => {
...
+                  <ButtonOutline
+                    className={clsx(styles.button, styles.buttonOutline)}
+                    text="Sign In"
+                  />
+                </Link>
+              </div>
+            )}
+            <div className={styles.searchWrapper}>
+              {isAuth && <SimpleAutoFilter />}
             </div>
```
oleksandr-danylchenko: Don't we need the?
```suggestion
           {isAuth && <div className={styles.searchWrapper}>
               <SimpleAutoFilter />}
            </div>}
```

Because why should an empty `div` hang in the DOM?

oleksandr-danylchenko (approved):

PR #159, opened 2022-08-31
Fix bug 24

setar92:
deleted hbs file. refactored send mail service.


backend/src/services/mail-verification/send-activation-link/send-link.ts
```diff
@@ -1,12 +1,13 @@
...
 import { ENV } from '@common/enums/app/app';
 import nodemailer from 'nodemailer';
-import hbs from 'nodemailer-express-handlebars';
 
 import { MailActivate } from './constants';
+
 const sendLink = async (email: string, token: string): Promise<void> => {
-  const templatePath = path.join(__dirname, 'templates');
+  const templatePath =
+    '@services/mail-verification/send-activation-link/templates/email-template.ts';
```
oleksandr-danylchenko, line 8: @setar92 
This won't work because only `tsc-alias` knows how to resolve such path

backend/src/services/mail-verification/send-activation-link/send-link.ts
```diff
@@ -1,12 +1,13 @@
...
+
+  const templateService = await import(templatePath);
```
oleksandr-danylchenko: Does the `tsc-alias` resolve such a path in dynamic import?

setar92: you are right, it's not.

backend/src/services/mail-verification/send-activation-link/send-link.ts
```diff
@@ -16,23 +14,12 @@ const sendLink = async (email: string, token: string): Promise<void> => {
...
-
   const options = {
     from: ENV.MAIL.FROM_EMAIL_VALIDATE,
     to: email,
     subject: MailActivate.SUBJECT,
     template: 'email',
+    html: templateService.getMessage(`${MailActivate.ACTIVATE_URL}${token}`),
     context: {
       link: `${MailActivate.ACTIVATE_URL}${token}`,
     },
```
oleksandr-danylchenko: Do we still need the `context`? I see that we already pass the `link` as the param into the `getMessage`

setar92: you are right, now we don't need context. 
deleted.

oleksandr-danylchenko (approved):

PR #160, opened 2022-08-31
like-button-lead-to-sign-in-screen

stanislav-ostr:


oleksandr-danylchenko (approved):

frontend/src/components/new-car-card/new-car-card.tsx
```diff
@@ -65,6 +70,12 @@ const NewCarCard: React.FC<ExtendedCarCardPropsType> = (props) => {
 
   const handleLikeClick = (event: React.MouseEvent): void => {
     event.stopPropagation();
+
+    if (!authToken) {
+      navigate('../sign-in');
+      return;
+    }
```
oleksandr-danylchenko: Please use the `AppRoute` enum instead of the `/sign-in` path

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #168, opened 2022-09-01
Add endpoints for comparisons

ashonuyak:
See - https://app.asana.com/0/1202525679663009/1202867677736072/f

### DIFF
--- a/backend/src/controllers/comparisons/comparisons.controller.ts
+++ b/backend/src/controllers/comparisons/comparisons.controller.ts
@@ -1,0 +1,95 @@
+import {
+  TypedRequestBody,
+  TypedRequestQuery,
+} from '@common/types/controller/controller';
+import { ComparisonType } from '@prisma/client';
+import * as comparisonsService from '@services/comparisons/comparisons.service';
+import { NextFunction, Response } from 'express';
+import httpStatus from 'http-status-codes';
+
+const addCarToComparison = async (
+  req: TypedRequestBody<{ complectationId: string; userId: string }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const comparison = await comparisonsService.addCarToComparison(req.body);
+    res.status(httpStatus.CREATED).json(comparison);
+  } catch (error) {
+    next(error);
+  }
+};
+
+const changeComparisonType = async (
+  req: TypedRequestBody<{
+    type: ComparisonType;
+    userId: string;
+  }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const comparison = await comparisonsService.changeComparisonType(req.body);
+    res.status(httpStatus.OK).json(comparison);
+  } catch (error) {
+    next(error);
+  }
+};
+
+const clearComparison = async (
+  req: TypedRequestBody<{
+    userId: string;
+  }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const comparison = await comparisonsService.clearComparison(
+      req.body.userId,
+    );
+    res.status(httpStatus.OK).json(comparison);
+  } catch (error) {
+    next(error);
+  }
+};
+
+const deleteCarFromComparison = async (
+  req: TypedRequestQuery<{
+    complectationId: string;
+    userId: string;
+  }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const comparison = await comparisonsService.deleteCarFromComparison(
+      req.query,
+    );
+    res.status(httpStatus.OK).json(comparison);
+  } catch (error) {
+    next(error);
+  }
+};
+
+const getActiveComparisonCars = async (
+  req: TypedRequestQuery<{ userId: string }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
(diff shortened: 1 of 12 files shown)


backend/src/services/comparisons/comparisons.service.ts
```diff
@@ -0,0 +1,194 @@
...
+    where: {
+      active: true,
+      user_id: userId,
+    },
+    select: {
+      id: true,
+    },
+  });
+  if (!activeComparison) {
+    throw new Error('There is no active comparison');
```
R0ixy: Maybe it will be better to return an empty array in case of no active comparison?  

If there isn't any, user still can visit comparison page and see nothing, so it looks like there isn't any reason to return  `400 Bad Request`.

backend/src/controllers/comparisons/comparisons.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+  req: TypedRequestBody<{
+    type: ComparisonType;
+    userId: string;
+  }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const comparison = await comparisonsService.changeComparisonType(req.body);
+    res.status(httpStatus.CREATED).json(comparison);
```
oleksandr-danylchenko: This should be the `PATCH` request and it should return `OK`

backend/src/controllers/comparisons/comparisons.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+    userId: string;
+  }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const comparison = await comparisonsService.clearComparison(
+      req.body.userId,
+    );
+    res.status(httpStatus.CREATED).json(comparison);
```
oleksandr-danylchenko: `PATCH` - `OK`

backend/src/controllers/comparisons/comparisons.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+    userId: string;
+  }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const comparison = await comparisonsService.deleteCarFromComparison(
+      req.query,
+    );
+    res.status(httpStatus.CREATED).json(comparison);
```
oleksandr-danylchenko: `DELETE` - `OK`

backend/src/controllers/comparisons/comparisons.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+const getActiveComparisonCars = async (
+  req: TypedRequestQuery<{ userId: string }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const comparison = await comparisonsService.getActiveComparisonCars(
+      req.query.userId.toString(),
+    );
+    res.status(httpStatus.CREATED).json(comparison);
```
oleksandr-danylchenko: What do we create on `GET`? :eyes:

backend/src/routes/comparisons/comparisons.router.ts
```diff
@@ -0,0 +1,14 @@
+import * as comparisonsController from '@controllers/comparisons/comparisons.controller';
+import { Router } from 'express';
+
+const PATH = '/comparisons';
+
+const comparisonsRouter = Router();
+
+comparisonsRouter.post(PATH, comparisonsController.addCarToComparison);
+comparisonsRouter.put(PATH, comparisonsController.changeComparisonType);
+comparisonsRouter.put(`${PATH}/clear`, comparisonsController.clearComparison);
```
oleksandr-danylchenko: `PATCH` because we change only one prop of the instance. `PUT` is usually used we're able to change the whole instance. 

This separation isn't 100% strict, but let's stick to it to consistency

backend/src/services/comparisons/comparisons.service.ts
```diff
@@ -0,0 +1,194 @@
...
+        comparison_id: comparison.id,
+        complectation_id: complectationId,
+        position: carsInComparisonAmount + 1,
+      },
+    });
+
+    return comparison;
+  }
+
+  if (!activeComparison) {
```
oleksandr-danylchenko: Why do we need this check? :slightly_smiling_face: (prev condition allows us to safely suggest that the `activeComparison` doesn't exist)

backend/src/services/comparisons/comparisons.service.ts
```diff
@@ -0,0 +1,194 @@
...
+    a.position > b.position ? 1 : -1,
+  );
+
+  return Promise.all(
+    sortedComplectations.map(async (car) => {
+      return await prisma.complectation.findUnique({
+        where: { id: car.complectation_id },
+      });
+    }),
+  );
```
oleksandr-danylchenko: Oh-oh-oh... There are multiple mistakes here. Can we have a huddle on it?

oleksandr-danylchenko: But I'll leave this for the future - https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#in

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #170, opened 2022-09-01
fix paths to the templates

ashonuyak:
See - https://app.asana.com/0/1202525679663009/1202781844236492/f

### DIFF
--- a/frontend/src/components/reset-password/reset-password.tsx
+++ b/frontend/src/components/reset-password/reset-password.tsx
@@ -21,7 +21,7 @@ export const ResetPassword = (): React.ReactElement => {
   const userId = queryParams.get('id') as string;
   useEffect(() => {
     if (!userId) {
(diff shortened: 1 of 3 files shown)


frontend/src/components/reset-password/reset-password.tsx
```diff
@@ -21,7 +21,7 @@ export const ResetPassword = (): React.ReactElement => {
   const userId = queryParams.get('id') as string;
   useEffect(() => {
     if (!userId) {
-      navigate(`/${AppRoute.NOT_FOUND}`);
+      navigate(`${AppRoute.NOT_FOUND}`);
```
oleksandr-danylchenko: Why did you wrap the string into the string interpolation? I believe you can use the `AppRoute.NOT_FOUND` directly

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #175, opened 2022-09-01
connect simple search with backend

brovarenko:
See - https://app.asana.com/0/1202525679663009/1202804889720773/f


frontend/src/store/found-car/slice.ts
```diff
@@ -0,0 +1,25 @@
...
+const { reducer, actions } = createSlice({
+  name: 'found-cars',
+  initialState,
+  reducers: {
+    setCars: (state, action: PayloadAction<FilterReturnType>) => {
+      state.cars = action.payload;
+    },
+  },
+  extraReducers: {},
+});
```
oleksandr-danylchenko: Why do we need a slice for the data that is already stored in the RTKQ cache?

frontend/src/store/found-car/slice.ts
```diff
@@ -0,0 +1,25 @@
...
+interface IFoundCars {
+  cars: FilterReturnType;
+}
+
+const initialState: IFoundCars = {
+  cars: [],
+};
+
+const { reducer, actions } = createSlice({
+  name: 'found-cars',
```
oleksandr-danylchenko: Please don't use `kebab-case` in the JS code. Such variables are not accessible with the `.` notation
```suggestion
  name: 'foundCars',
```

oleksandr-danylchenko (changes requested):

PR #181, opened 2022-09-02
Smooth scrolling to top after search click

elador13:
scroll-top

### DIFF
--- a/frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
+++ b/frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
@@ -57,6 +57,10 @@ const AdvancedAutoFilter: FC<AdvancedAutoFilterProps> = (props) => {
 
   const doSearch = (): void => {
     search(queryParams);
+    window.scrollTo({
+      top: 0,
+      behavior: 'smooth',
+    });
   };
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


elador13:
P.S A little freezes when scrolling - it's my overloaded laptop(

frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
```diff
@@ -57,6 +58,7 @@ const AdvancedAutoFilter: FC<AdvancedAutoFilterProps> = (props) => {
 
   const doSearch = (): void => {
     search(queryParams);
+    scroll.scrollToTop({ duration: 700 });
```
oleksandr-danylchenko: I don't think that we need a separate dependency for just the scrolling.
We can simply use the build-it [`window.scroll`](https://developer.mozilla.org/en-US/docs/Web/API/Window/scroll) method

elador13: @oleksandr-danylchenko because "smooth" scrolling doesn't work in `window.scroll({
      top: 0,
      left: 0,
      behavior: "smooth"
    })`

oleksandr-danylchenko: Whoa, no way :thinking:  
image

oleksandr-danylchenko: Could you try the `scrollTo` method?

Aleksandr-Shulha-dev (approved):

oleksandr-danylchenko (approved):

PR #183, opened 2022-09-02
Fixed header position to static

AlexanderKhapchenko:


iryb:
Why not `position: sticky`? In this case Header always will be on top of the page but won't overlap the page content.

AlexanderKhapchenko:
@iryb Oleksandr Danylchenko, suggested making it static

see
https://binary-academy-2022.slack.com/archives/C03SA9P93C6/p1662052080006249?thread_ts=1662049214.550399&cid=C03SA9P93C6

oleksandr-danylchenko (approved):

PR #184, opened 2022-09-02
Fix/logout button

AlexanderKhapchenko:
Fixed the width of the logout button and added a temporary solution for functionality

### DIFF
--- a/frontend/src/components/header/header.tsx
+++ b/frontend/src/components/header/header.tsx
@@ -5,7 +5,7 @@ import Logo from '@assets/images/logo.svg';
 import { AppRoute } from '@common/enums/app/app-route.enum';
 import { EditProfile } from '@components/edit-profile/edit-profile';
 import { UnauthorisedElements } from '@components/header/unauthorised-elements/unauthorised-elements';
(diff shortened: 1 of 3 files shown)


frontend/src/components/header/header.tsx
```diff
@@ -82,7 +83,12 @@ export const Header = (): React.ReactElement => {
     },
     logout: {
       label: 'Logout',
-      onClick: () => navigate('#'),
+      onClick: (): void => {
+        logOut();
```
R0ixy: If I'm not mistaken your `logOut` does completely nothing because we need to call `dispatch(logOut())`

frontend/src/components/header/header.tsx
```diff
@@ -82,7 +83,12 @@ export const Header = (): React.ReactElement => {
...
+        //A temporary solution until logout is implemented
+        localStorage.removeItem('token');
+        location.reload();
```
R0ixy: This functionality was already implemented as a middleware for reducer `logOut` in PR 151

RomanMahola (approved):

PR #186, opened 2022-09-02
fix redirect link

ashonuyak:


backend/src/services/auth/auth.service.ts
```diff
@@ -101,7 +101,7 @@ const requestPasswordReset = async (email: string): Promise<string> => {
     data: { password_change_token: resetToken },
   });
 
-  const link = `${ENV.APP.SERVER_HOST}:${ENV.APP.SERVER_PORT}${ENV.API.V1_PREFIX}/auth/local/reset-password-check-token?token=${resetToken}`;
+  const link = `${process.env.VITE_API_ORIGIN_URL}/auth/local/reset-password-check-token?token=${resetToken}`;
```
oleksandr-danylchenko: Are you sure you want to access the `VITE_` variable in the backend? :sweat_smile:

ashonuyak: @oleksandr-danylchenko  I just need the user redirected to the server. Is it the wrong way? Also i did it the way it's done in send-mail method for mail verification

RomanMahola: What if your frontend does not know anything about backend.
I try to explain.
Imagine the situation that your frontend server is in Dubai and backend location is Chicago. They easily can communicate with each other (yep, with delay, but still can). We can migrate our services any time. So after migration you just wouldn't have any access to frontend URL. 

Long story short. It's ok to add another variable to `.env`

PR #193, opened 2022-09-02
Sticky scrollable filter form + fix reset filters button

elador13:
See - https://app.asana.com/0/1202525679663009/1202893629985159/f
See - https://app.asana.com/0/1202525679663009/1202914210974976/f

The code is larger than I expected, because we need to set the initial height based on the current scroll position (User can reload page in different scroll positions, not only in the top of page).

**Result:**

https://user-images.githubusercontent.com/49813216/188244001-234881ec-5d40-44fa-b75a-e226d5b29231.mp4

**Page reloading in different scroll positions:**

https://user-images.githubusercontent.com/49813216/188244069-22543918-0810-46aa-82e4-872dc8981cb5.mp4


frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
```diff
@@ -64,6 +72,31 @@ const AdvancedAutoFilter: FC<AdvancedAutoFilterProps> = (props) => {
     });
   };
 
+  const [filterContainer, setRef] = useState<HTMLDivElement | null>(null);
```
oleksandr-danylchenko: `setRef` -> `setFilterContainerRef`

frontend/src/helpers/utils/get-element-height.ts
```diff
@@ -0,0 +1,14 @@
...
+  elementHeight +=
+    parseInt(window.getComputedStyle(element).getPropertyValue('margin-top')) ||
+    0;
+  elementHeight +=
+    parseInt(
+      window.getComputedStyle(element).getPropertyValue('margin-bottom'),
+    ) || 0;
+
+  return elementHeight;
+};
```
oleksandr-danylchenko: Can you add some explanation? Because it's not that intuitive why we should consider the margins :thinking:

frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
```diff
@@ -64,6 +72,31 @@ const AdvancedAutoFilter: FC<AdvancedAutoFilterProps> = (props) => {
...
+
+  useLayoutEffect(() => {
+    if (!filterContainer) return;
+
+    const headerHeight =
+      document.querySelector('#mainHeader')?.clientHeight || 0;
+
+    const searchTitleHeight = getElementHeight(
+      document.querySelector('#searchTitle'),
+    );
```
oleksandr-danylchenko: Why in the first case you just refer to the `clientHeight` and in the second one you use the `getElementHeight`. Because semantically they do the same thing.
If you would like to convey that `getElementHeight` does smth specific - use more specific and explicit name

frontend/src/helpers/utils/get-element-height.ts
elador13: Because "Search" title has top,bottom margins, and without them, title height is only 32.2px, but the real height is 92px

image

frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
elador13: Ok, I can change name -> getElementHeightWithMargins

frontend/src/helpers/utils/get-element-height.ts
oleksandr-danylchenko: До речі... У JS справді немає способу отримати висоту елементу включно з його margin'ами. Але варто бути уважним та зважати що `getElementHeightWithMargins` може поломатися якщо висота елемента вказана явно (JS такий JS :clown_face:)

Demo - https://playcode.io/954066/

oleksandr-danylchenko (approved):
https://user-images.githubusercontent.com/68850090/188266818-d6081656-3b54-4af3-af1f-a4d7f3b95550.mp4

frontend/src/helpers/utils/get-element-height.ts
elador13: Yeah, also this "separately scrolling" feature strongly depends of html elements above. And after every adding/deleting new htmls I need to add/delete their height in useLayoutEffect to correctly calculate formTop position. I hope it'll be rarely...

PR #194, opened 2022-09-03
fix autoria request

R0ixy:
Now we use `compectation_Ids` to get `autoria_codes` and make request

### DIFF
--- a/backend/src/controllers/cars/cars.controller.ts
+++ b/backend/src/controllers/cars/cars.controller.ts
@@ -11,7 +11,7 @@ import { NextFunction, Request, Response } from 'express';
 
 const carsSearchAutoria = async (
   req: TypedRequestQuery<{
-    complectationId: string[];
+    complectationId: string;
     page: string;
   }>,
   res: Response,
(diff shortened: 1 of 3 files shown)


backend/src/controllers/cars/cars.controller.ts
```diff
@@ -10,12 +10,18 @@ import * as carsService from '@services/cars/cars.service';
...
 
 const carsSearchAutoria = async (
-  req: TypedRequestQuery<CarsSearchParams>,
+  req: TypedRequestQuery<{
+    complectationId: string[];
+    page: string;
+  }>,
   res: Response,
   next: NextFunction,
 ): Promise<void> => {
```
oleksandr-danylchenko: I have a few questions: 
1. `complectationId` is singular, but its type is array :eyes:?
2. Why do we allow to pass multiple complectations at once? We will show autoria options for only one complication at a time

R0ixy: On the page where we can choose complectations, we use checkbox so it looks like we can specify many complactations

oleksandr-danylchenko: Nope, the checkbox was removed from the design and implementation plan. It was discussed in the meetings.

Also, on e-katalog the checkbox is responsible for adding an item to the comparison

oleksandr-danylchenko (approved):
Nice! Thanks :raised_hands:

PR #198, opened 2022-09-04
Add liking functionality to search list

brovarenko:
See - https://app.asana.com/0/1202525679663009/1202897278017237/f
like


backend/src/services/preferences/wishlist.service.ts
```diff
@@ -66,6 +66,25 @@ const deleteWishlist = async (input: WishlistInput): Promise<void> => {
...
+      OR: [
+        {
+          model_id: modelId,
+        },
+        {
+          complectation_id: complectationId,
+        },
+      ],
+    },
+  });
```
oleksandr-danylchenko: We do the same query on line 15 and in other methods. Could you please extract them into a reusable function?

backend/src/services/preferences/wishlist.service.ts
```diff
@@ -66,6 +66,25 @@ const deleteWishlist = async (input: WishlistInput): Promise<void> => {
...
+  const result = Boolean(wishlist?.model_id || wishlist?.complectation_id);
+  return result;
```
oleksandr-danylchenko:
```suggestion
  return Boolean(wishlist?.model_id || wishlist?.complectation_id);
```

frontend/src/components/car-list-item/car-list-item.tsx
```diff
@@ -1,41 +1,79 @@
...
+  useDeleteWishlistMutation,
+} from '@store/queries/preferences/wishlist';
 import { clsx } from 'clsx';
 import { Swiper, SwiperSlide } from 'swiper/react';
 
 import styles from './styles.module.scss';
 
 const CarListItem: React.FC<CarListItemProps> = (props) => {
   const { model_id, complectations_id } = props;
+  const authToken = useAppSelector((state) => state.auth.token);
```
oleksandr-danylchenko: Could you please select a shared selector for the token? I saw that it's extracted like this in a few places

frontend/src/components/car-list-item/car-list-item.tsx
```diff
@@ -1,41 +1,79 @@
...
+  };
+
+  const handleLikeClick = (event: React.MouseEvent): void => {
+    event.stopPropagation();
+    const data: WishlistInput = { modelId: model_id };
+
+    if (!authToken) {
+      navigate(AppRoute.SIGN_IN);
+      return;
+    }
```
oleksandr-danylchenko: This should be handled in some shared place. Like a RTK middleware. Because we have the same exact code for the likes on the cards

frontend/src/components/car-list-item/car-list-item.tsx
```diff
@@ -1,41 +1,79 @@
...
+
+    isLikedModel ? handleDeleteWishlist(data) : handleCreateWishlist(data);
+  };
```
oleksandr-danylchenko: This code is fine for now. But I believe it will be replaced with the shared context on which @iryb is working on right now.

frontend/src/components/car-list-item/styles.module.scss
```diff
@@ -108,3 +108,37 @@
...
+  }
+
+  &.isLiked,
+  &:hover {
+    path {
+      transition: all 300ms;
+      fill: var(--blue100);
+    }
+  }
+}
```
oleksandr-danylchenko: Don't we already have the styles for the like button?

oleksandr-danylchenko (changes requested):

frontend/src/components/car-list-item/styles.module.scss
brovarenko: I added common component llike button

frontend/src/components/common/like-button/like-button.tsx
```diff
@@ -0,0 +1,24 @@
+import { LikeButtonPropsType } from '@common/types/types';
+import { clsx } from 'clsx';
+
+import { HeartIcon } from '../icons/icons';
+import styles from './styles.module.scss';
+
+const LikeButtton: React.FC<LikeButtonPropsType> = ({ isLiked, onClick }) => {
```
oleksandr-danylchenko: Props must be declared before the component in the same file

frontend/src/components/common/like-button/like-button.tsx
```diff
@@ -0,0 +1,24 @@
...
+        className={clsx(
+          styles.button,
+          styles.iconButton,
+          isLiked && styles.isLiked,
+        )}
+        onClick={onClick}
+      >
+        <HeartIcon />
+      </button>
+    </>
```
oleksandr-danylchenko: Why do you use `React.Fragment` here? `button` can be immediately returned

frontend/src/store/middlewares/middlewares.ts
```diff
@@ -0,0 +1,27 @@
...
+  effect: (action, listenerApi) => {
+    const token = listenerApi.getState().auth.token;
+    if (!token) {
+      window.location.replace(AppRoute.SIGN_IN);
+      return;
+    }
+  },
+});
+
+export { checkTokenMiddleware };
```
oleksandr-danylchenko: Why it's in the `middlewares.ts` file? Shouldn't it be in more specific file?

frontend/src/components/car-list-item/car-list-item.tsx
```diff
@@ -37,6 +42,15 @@ const CarListItem: React.FC<CarListItemProps> = (props) => {
...
   const [modelTransmission, setModelTransmission] = useState<string>();
 
+  const { likedCars, handleLikeClick } = useContext(WishlistContext);
+  const isLiked = likedCars?.includes(model_id);
+
+  const likeClick = (event?: React.MouseEvent): void => {
+    event?.stopPropagation();
+    const data: WishlistInput = { modelId: model_id };
+    handleLikeClick(data);
+  };
```
oleksandr-danylchenko, line 52: Where do you get the `model_id`? And why do we have variables in `snake_case` in code?

oleksandr-danylchenko (changes requested):

frontend/src/components/car-list-item/car-list-item.tsx
brovarenko: We take this data from props, it comes to props in the current form from the backend.

PR #200, opened 2022-09-04
Fixed bug 36 (added validation unique email for editing user by admin)

o-rudych:
https://app.asana.com/0/1202525679663009/1202912993210480/f

### DIFF
--- a/backend/src/middlewares/users/edit-user.middleware.ts
+++ b/backend/src/middlewares/users/edit-user.middleware.ts
@@ -1,0 +1,30 @@
+import { ErrorMessage, editUserSchema } from '@autoline/shared';
+import { TypedRequest } from '@common/types/controller/controller';
+import { UserUpdateInput } from '@controllers/users/users.controller';
+import { prisma } from '@data/prisma-client';
+import { NextFunction, Response } from 'express';
+import httpStatus from 'http-status-codes';
+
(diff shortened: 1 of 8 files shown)


backend/src/middlewares/users/edit-user.middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+    });
+    if (existedUser) next(new Error('User with this email already exists!'));
+
+    next();
+  } catch (err) {
+    const { message } = err as ErrorMessage;
+    res.status(httpStatus.FORBIDDEN).json({ error: message });
+    next(err);
+  }
+};
```
oleksandr-danylchenko: This should be in the service layer, not in the middleware

oleksandr-danylchenko (changes requested):

backend/src/middlewares/users/edit-user.middleware.ts
o-rudych: All this code to the service layer?
Or may be `await editUserSchema.validate(req.body);` should be left here in middleware? I assume it.

o-rudych: I've moved this part of code.

oleksandr-danylchenko (approved):

PR #201, opened 2022-09-04
Fixed bug 37 (ability to save empty optional fields on edit user form by admin)

o-rudych:
https://app.asana.com/0/1202525679663009/1202912993210493/f

### DIFF
--- a/shared/src/validation-schemas/edit-user/edit-user.validation-schema.ts
+++ b/shared/src/validation-schemas/edit-user/edit-user.validation-schema.ts
@@ -25,7 +25,7 @@ const editUserSchema = Yup.object().shape({
   }),
   location: Yup.string()
     .nullable()
-    .transform((value) => (value ? value : null))
+    .transform((value) => value || null)
     .min(2, 'Location must be at least 2 characters')
     .max(150, 'Location must not exceed 150 characters')
     .matches(
--- a/shared/src/validation-schemas/edit-user/edit-user.validation-schema.ts
+++ b/shared/src/validation-schemas/edit-user/edit-user.validation-schema.ts
@@ -19,8 +19,13 @@ const editUserSchema = Yup.object().shape({
     .matches(/@[A-Za-z0-9.\-_]*$/, UserValidationMessage.INVALID_EMAIL)
     .matches(/^[\S]{1,35}@/, UserValidationMessage.INVALID_EMAIL_LENGTH)
     .matches(/@[\S]{3,35}$/, UserValidationMessage.INVALID_EMAIL_LENGTH),
-  phone: Yup.string().matches(phoneReg, 'Phone is invalid'),
+  phone: Yup.string().matches(phoneReg, {
+    message: 'Phone is invalid',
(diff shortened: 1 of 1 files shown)


shared/src/validation-schemas/edit-user/edit-user.validation-schema.ts
```diff
@@ -19,8 +19,13 @@ const editUserSchema = Yup.object().shape({
...
     .matches(/^[\S]{1,35}@/, UserValidationMessage.INVALID_EMAIL_LENGTH)
     .matches(/@[\S]{3,35}$/, UserValidationMessage.INVALID_EMAIL_LENGTH),
-  phone: Yup.string().matches(phoneReg, 'Phone is invalid'),
+  phone: Yup.string().matches(phoneReg, {
+    message: 'Phone is invalid',
+    excludeEmptyString: true,
+  }),
   location: Yup.string()
+    .nullable()
+    .transform((value) => (value ? value : null))
```
oleksandr-danylchenko: `value ? value : null` === `value || null`

Why do we need the `null` value specifically?

o-rudych: @oleksandr-danylchenko, I need to validate string to be empty or to fit next validation rules.
So I've decided to do it this way.
If there is a better way to apply validation rules only to a non-empty string, I'd appreciate some advice.

oleksandr-danylchenko: I don't fully understand the problem :thinking: 
What will happen if you just don't make the field required with `.required`? Why do we even need the `nullable`?

o-rudych: If we don't use `.required` but use `.min(2, 'Location must be at least 2 characters')`, we'll ger error `"Location must be at least 2 characters"` in case empty field "Location". We can check it now on our staging.

oleksandr-danylchenko: Okay... 
Then could help me to understand that the `transform` does here and how it helps the `nullable`? (I really haven't worked with Yup for quite some time and it's not obvious for me :slightly_smiling_face:)

o-rudych: I am also using this method for the first time)
`.nullable()` makes the value to accept null, `transform` callback returns the value as null if value is empty
https://github.com/jquense/yup/issues/1086#issuecomment-988143675

oleksandr-danylchenko: Thanks for sharing the link :+1:. Now it makes sense to me. But I still think that it can be simplified just to the `value || null`

o-rudych: @oleksandr-danylchenko, good suggestion, thanks! Done.

oleksandr-danylchenko (approved):

PR #203, opened 2022-09-05
bug_41: Fix display of car card in 'New Cars' section (mobile layout)

stanislav-ostr:


Aleksandr-Shulha-dev (commented):
If we remove the property "justify-items: center" for ".secondContainerCards", this bug disappears. I think the problem is this.

Aleksandr-Shulha-dev:
https://gyazo.com/7f0f61c5b4c251b73d7b236ac66a8dce

stanislav-ostr:
There is a requirement to place content in the center in the mobile version
Снимок экрана 2022-09-05 161659

Aleksandr-Shulha-dev:
> There is a requirement to place content in the center in the mobile version Снимок экрана 2022-09-05 161659

This happens because the container has a right margin. The card width equals the width of the container if we delete "justify-items". In order for the cards to be in the center, you need to reduce margin rigth for container.

image

oleksandr-danylchenko:
@Aleksandr-Shulha-dev @stanislav-ostr 
To center the content w/o flex you can specify right and left margins as `auto`. Or if you want to work with flex, try to use `flex: 1` on the card item. I'll prevents its shrinking to take just enough space to fit the content

PR #209, opened 2022-09-05
Add comparison counter logic frontend

R0ixy:
Asana - https://app.asana.com/0/1202525679663009/1202867677736076/f


backend/src/services/comparisons/comparisons.service.ts
```diff
@@ -176,10 +176,25 @@ const getActiveComparisonCars = async (
...
+      complectation_id: true,
+    },
+    where: {
+      comparison: {
+        user_id: userId,
+      },
+    },
+  });
+  return comparison.map((c) => c.complectation_id);
+};
```
oleksandr-danylchenko: `comparison` singular but contains an array

backend/src/controllers/comparisons/comparisons.controller.ts
```diff
@@ -54,31 +64,47 @@ const clearComparison = async (
...
+  } catch (error) {
+    next(error);
+  }
+};
+
+const getActiveComparisonStatus = async (
+  req: TypedRequestBody<{ tokenPayload: TokenPayload }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
```
oleksandr-danylchenko, line 104: Get requests cannot have the `body`. Is the `tokenPayload` injected in the middleware?

oleksandr-danylchenko (changes requested):

backend/src/controllers/comparisons/comparisons.controller.ts
R0ixy: Yes it is injected

oleksandr-danylchenko: Hmm, then it's injected in the wrong place. Because `tokenPayload` should be a part of the request, **_not its body_**. Could create a new type that will include the `tokenPayload` in the request but still would be able to have a generic type for the body/query?

R0ixy: Yeah, sure. I think we need to fix it in all controllers so I'll make separate PR for it

oleksandr-danylchenko: Good, then lets defer it for now :ok:

oleksandr-danylchenko (approved):

PR #212, opened 2022-09-06
Aws s3

RomanMahola:


.github/workflows/deploy-stage.yml
```diff
@@ -80,6 +80,24 @@ jobs:
...
+          flags: --recursive
+
+      - name: Copy secret files to server
+        uses: appleboy/scp-action@master
+        with:
+          host: ${{ secrets.SSH_HOST }}
+          username: ${{ secrets.SSH_USERNAME }}
+          key: ${{ secrets.SSH_KEY }}
+          source: .env
+          target: app
```
oleksandr-danylchenko, line 99: Why don't we copy env files from S3 directly to the `app` directory?

RomanMahola: We can, but will we have some advances?)

PR #215, opened 2022-09-06
Timer to hide the notification

Aleksandr-Shulha-dev:
https://gyazo.com/d349580f501dbc6b33b08717541f57a5

### DIFF
--- a/frontend/src/components/common/notification/notification.tsx
+++ b/frontend/src/components/common/notification/notification.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react';
+import React, { FC, useEffect } from 'react';
 
 import { ButtonOutline } from '../button-outline/button-outline';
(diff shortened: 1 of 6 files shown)


frontend/src/components/common/notification/notification.tsx
```diff
@@ -16,6 +16,8 @@ const Notification: FC<ModalProps> = (props) => {
 
   const closeNotification = (): void => setIsOpen(!isOpen);
 
+  isOpen && setTimeout(closeNotification, 6000);
```
oleksandr-danylchenko: All the side effects must be declared and cleared inside the `useEffect`. Because now on each rerender a new timeout will be created

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #217, opened 2022-09-06
Fixed fields names in edit profile

AlexanderKhapchenko:


o-rudych (commented):
I guess it's better to change the "Email" label to "E-mail" as well.

oleksandr-danylchenko (approved):

PR #218, opened 2022-09-06
Fix token payload logic

R0ixy:
Fixed `tokenPayload` logic as was discussed here https://github.com/BinaryStudioAcademy/bsa-2022-autoline/pull/209#discussion_r963465394

Not sure if it's perfect solution because now we need to cast `tokenPayload` type everywhere

Also decided to delete `test-protected-route` as it was created only for testing and we don't need it anymore.


backend/src/common/types/controller/controller.ts
```diff
@@ -1,26 +1,32 @@
...
 
-interface TypedRequest<T extends Query, U> extends Request {
+interface TypedRequest<T extends Query, U> extends AuthRequest {
   body: U;
   query: T;
 }
 
+interface AuthRequest extends Request {
+  tokenPayload?: TokenPayload;
+}
```
oleksandr-danylchenko, line 24: Make the `tokenPayload` defined. Then devs will be able access it w/o `as` in places where it's expected

oleksandr-danylchenko (changes requested):

backend/src/common/types/controller/controller.ts
oleksandr-danylchenko: This can be done in a simpler way and w/o `as`:

```ts
type AuthTypedRequestBody<T> = TypedRequestBody<T> & AuthRequest;

type AuthTypedRequestQuery<T extends Query> = TypedRequestQuery<T> &
  AuthRequest;

type AuthTypedRequestParams<T extends ParamsDictionary> =
  TypedRequestParams<T> & AuthRequest;

type AuthTypedRequestParams<T extends Query, U> = TypedRequest<T, U> &
  AuthRequest;

interface AuthRequest extends Request {
  tokenPayload: TokenPayload;
}
```

oleksandr-danylchenko (changes requested):

backend/src/common/types/controller/controller.ts
R0ixy: We still have problems with types. If we use `req: AuthRequest` in role.middleware we will get TypeError in routers. I tried `Omit<>`, `Partial<>` and so on. Nothing helps but optional `tokenPayload?` in `AuthRequest`.

Maybe there is a way to specify one type as the input to middleware and another as the output?

oleksandr-danylchenko: Where exactly do you try to use the `req: AuthRequest`?

R0ixy: In `role.middleware`. We pass plain request and need to add tokenPayload to it

backend/src/controllers/comparisons/comparisons.controller.ts
```diff
@@ -91,13 +90,13 @@ const getActiveComparisonCars = async (
...
-  req: AuthRequest,
+  req: Request,
   res: Response,
   next: NextFunction,
 ): Promise<void> => {
   try {
     const comparison = await comparisonsService.getActiveComparisonStatus(
-      req.tokenPayload?.sub as string,
+      req.tokenPayload.sub,
     );
```
oleksandr-danylchenko: In controllers, we should use typed requests with typed body!. Untyped `Request` can be used only in middelwares

R0ixy:
I'm not sure why all these previous commits added here after merge...

backend/src/controllers/comparisons/comparisons.controller.ts
R0ixy: Ouch, but now we don't have any difference between them

oleksandr-danylchenko:
@R0ixy 
Please change the target branch to the `fix-token-payload-logic`
image

R0ixy:
I accidentally made a rebase so I think few cherry-picks and force push is better for this case

backend/src/common/types/controller/controller.ts
```diff
@@ -18,9 +28,28 @@ interface TypedRequest<T extends Query, U> extends Request {
...
+
+type AuthTypedRequestBody<T> = TypedRequestBody<T> & AuthRequest;
+
+type AuthTypedRequestQuery<T extends Query> = TypedRequestQuery<T> &
+  AuthRequest;
+
+type AuthTypedRequestParams<T extends ParamsDictionary> =
+  TypedRequestParams<T> & AuthRequest;
+
+type AuthTypedRequest<T extends Query, U> = TypedRequest<T, U> & AuthRequest;
```
oleksandr-danylchenko: Are these types still needed? Can't we just use the `TypedRequest...` as the `Request` already contains the `tokenPayload`

R0ixy: But... then what did you mean by this?  
https://github.com/BinaryStudioAcademy/bsa-2022-autoline/pull/218#discussion_r964528131

oleksandr-danylchenko: I meant that we must type the controller requests. But since then you moved `tokenPayload` to the global `Request` and we don't need to have additional `AuthTyped...` types anymore. Because of the existing `TypedRequest...` extends the `Request`

oleksandr-danylchenko (approved):

PR #223, opened 2022-09-07
Where buy backend

setar92:
Backend side for where-to-buy component.
 - Added helper for request to autoria by advert id.


backend/src/common/types/where-to-buy/where-to-buy.params.ts
```diff
@@ -0,0 +1,18 @@
...
+      last_id: number;
+    };
+  };
+}
+
+type WhereBuyRequestQuery = {
+  page: string;
+  id: string;
+  countpage: string;
+};
```
oleksandr-danylchenko, line 16: This also should be `interface`

backend/src/common/types/where-to-buy/where-to-buy.params.ts
```diff
@@ -0,0 +1,18 @@
...
+      count: number;
+      last_id: number;
+    };
+  };
+}
+
+type WhereBuyRequestQuery = {
+  page: string;
+  id: string;
+  countpage: string;
```
oleksandr-danylchenko, line 15: It's a good practice to put props in the types from more-> less specific order. In this case, `id` is the most specific field possible and should be first

backend/src/controllers/where-to-buy/where-to-buy.ts
```diff
@@ -0,0 +1,36 @@
...
+    const page = req.query.page;
+    const complectationId = req.query.id;
+    const countpage = +req.query.countpage;
+    const querysForRequest = await carsSearchAutoria(
+      complectationId,
+      page,
+      countpage,
+    );
+    const allAdverts = await getCarsAutoRia(querysForRequest);
+    const advertsId = allAdverts?.result.search_result.ids;
```
oleksandr-danylchenko, line 26: `advertsIds` would be better

backend/src/controllers/where-to-buy/where-to-buy.ts
```diff
@@ -0,0 +1,36 @@
...
+    const advertsInfo = advertsId?.map((car) => getCarByIdRia(car));
+    Promise.all(advertsInfo as readonly unknown[]).then((result) =>
+      res.json(result),
+    );
```
oleksandr-danylchenko, line 30:
```suggestion
    const advertsInfo = await Promise.all(advertsIds.map((car) => getCarByIdRia(car)))
    res.json(advertsInfo),
```

backend/src/controllers/where-to-buy/where-to-buy.ts
oleksandr-danylchenko: Also, how the `advertsIds` can be undefined

oleksandr-danylchenko (changes requested):

PR #225, opened 2022-09-07
add comparisons-top-table-bar functionality

ashonuyak:
See - https://app.asana.com/0/1202525679663009/1202919499021485/f

Appearance:
image

https://gyazo.com/7b92cae9c77399dea02ea28b5ead53e9
https://gyazo.com/a617d758c3e0efd91ed6891dd970db49
https://gyazo.com/ebda21a26062151f620de166adf5a7f4


frontend/src/components/comp-top-table-bar/components/complectation.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const [deleteWishlist] = useDeleteWishlistMutation();
+  const [deleteCar] = useDeleteCarFromComparisonMutation();
+
+  const handleCreateWishlist = async (data: WishlistInput): Promise<void> => {
+    await createWishlist(data);
+  };
+
+  const handleDeleteWishlist = async (data: WishlistInput): Promise<void> => {
+    await deleteWishlist(data);
+  };
```
iryb: Please, use context for like functionality. We moved these functions to the single place. Here is `car-list-item` component for the reference
https://github.com/BinaryStudioAcademy/bsa-2022-autoline/pull/198/files#

ashonuyak: damn thats nice

oleksandr-danylchenko:
@ashonuyak 
> react-beautiful-dnd

## I explicitly stated on the call that we won't use drag-&-drop! Please remove that library

backend/src/controllers/comparisons/comparisons.controller.ts
```diff
@@ -129,6 +129,38 @@ const getComparisonGeneralInfo = async (
...
+  req: TypedRequestBody<{ tokenPayload: TokenPayload }>,
+  res: Response,
+  next: NextFunction,
+): Promise<void> => {
+  try {
+    const comparisonPreview =
+      await comparisonsService.getActiveComparisonCarsPreview(
+        req.body.tokenPayload.sub,
+      );
+    res.status(httpStatus.OK).json(comparisonPreview);
```
oleksandr-danylchenko: Why preview? We don't preview anything :eye: We just show the cars in comparison

frontend/package.json
```diff
@@ -26,11 +26,13 @@
     "@sentry/react": "^7.9.0",
     "@sentry/tracing": "^7.9.0",
     "@types/node": "^17.0.45",
+    "@types/react-beautiful-dnd": "^13.1.2",
     "@types/react-router-hash-link": "^2.4.5",
     "@types/yup": "^0.29.14",
     "async-mutex": "^0.3.2",
     "clsx": "1.2.0",
     "react": "18.2.0",
+    "react-beautiful-dnd": "^13.1.1",
```
oleksandr-danylchenko: There shouldn't be DnD

frontend/src/components/comp-top-table-bar/components/complectation.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    setIsDeleted(true);
+    handleDeleteCar(props.car.id);
+  };
+
+  return (
+    <>
+      {!isDeleted && (
+        <Draggable
+          key={props.car.id}
```
oleksandr-danylchenko: You must use the guard clause here! Ask @iryb if you're not sure what it is

frontend/src/components/comp-top-table-bar/components/complectations-column.tsx
```diff
@@ -0,0 +1,52 @@
...
+        </div>
+      ) : (
+        <div className={styles.noCarsError}>
+          You have got no cars
+          <br />
+          in this comparison yet.
+        </div>
+      )}
+    </>
+  );
```
oleksandr-danylchenko: You can use guard clause here as well!

oleksandr-danylchenko (changes requested):

frontend/src/store/queries/comparisons/index.ts
```diff
@@ -53,6 +53,21 @@ export const comparisonsApi = api.injectEndpoints({
       query: () => `${API.COMPARISONS}/general`,
       providesTags: ['Comparisons'],
     }),
+    getComparisonCars: builder.query<ComparisonInfo[], void>({
+      query: () => ({
+        url: API.COMPARISONS,
+        method: 'GET',
+        providesTags: ['Comparisons'],
```
oleksandr-shulha-dev: I pulled your code to my branch and noticed some mistake. In this code you need added "providesTags" field after query method.

frontend/src/components/comp-top-table-bar/comp-top-table-bar.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import { Comparison } from './components/comparison';
+import { NoActiveComparison } from './components/no-active-comparison';
+import styles from './components/styles.module.scss';
+
+export const CompTopTableBar = (): React.ReactElement => {
+  const { data, isLoading } = useGetComparisonCarsQuery();
+  const [clearTable] = useClearComparisonMutation();
+  const [isCleared, setIsCleared] = useState(false);
+  const ref = useRef<HTMLInputElement>();
```
oleksandr-danylchenko: Use a more explicit name instead of the `ref`. And add some space between variables

frontend/src/components/comp-top-table-bar/comp-top-table-bar.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const cars = data?.slice().sort((a, b) => a.position - b.position);
+  const ids = cars?.map((car) => car.id);
```
oleksandr-danylchenko: `ids` name is too generic.

`.slice` and `.map` always generate new arrays! New arrays will be created on each rerender. Wrap then in `useMemo` if needed

frontend/src/components/comp-top-table-bar/comp-top-table-bar.tsx
```diff
@@ -0,0 +1,106 @@
...
+            });
+          };
+          currentElement.addEventListener('wheel', onWheel);
+          return (): void =>
+            currentElement.removeEventListener('wheel', onWheel);
+        }
+      },
+      { passive: false },
+    );
+  });
```
oleksandr-danylchenko: Memory leak!

Add handled into a function inside the `useEffect` and remove the listener on unmount

frontend/src/components/comp-top-table-bar/comp-top-table-bar.tsx
```diff
@@ -0,0 +1,106 @@
...
+              cars = undefined;
+            }
+
+            return (
+              <Comparison cars={cars} positions={initialData.carsPositions} />
+            );
+          })()
+        ) : (
+          <NoActiveComparison />
+        )}
```
oleksandr-danylchenko:
```
   ((): React.ReactElement => {
            let cars;
            const fetchedData = initialData.cars;
            if (fetchedData) {
              cars = initialData.carsPositions?.map((carId) => {
                const index = fetchedData.findIndex((car) => car.id === carId);
                return fetchedData[index];
              });
            } else {
              cars = undefined;
            }

            return (
              <Comparison cars={cars} positions={initialData.carsPositions} />
            );
```

Should be in a separate component or before the `render` method!

oleksandr-danylchenko (changes requested):

frontend/src/components/comp-top-table-bar/comp-top-table-bar.tsx
R0ixy: It looks like we don't need this functionality at all here. Because we gonna make all tables scroll in sync so scroll handling might be implemented outside of this component. 

It's not the part of this task I guess

frontend/src/components/comp-top-table-bar/comp-top-table-bar.tsx
```diff
@@ -17,13 +17,15 @@ export const CompTopTableBar = (): React.ReactElement => {
...
+  const cars = useMemo(
+    () => data?.slice().sort((a, b) => a.position - b.position),
+    [data],
+  );
+  const carsIds = cars?.map((car) => car.id);
   const initialData = {
     cars,
-    carsPositions: ids,
+    carsPositions: carsIds,
   };
```
oleksandr-danylchenko: You should use `useMemo` for `carsIds` too. Because new array will be created on each render

backend/src/services/comparisons/comparisons.service.ts
```diff
@@ -298,6 +314,139 @@ const getComparisonOptions = async (optionType: Type): Promise<string[]> => {
...
+    await prisma.comparisons_Complectations.update({
+      where: {
+        comparison_id_complectation_id: {
+          comparison_id: activeComparison.id,
+          complectation_id: complectationId,
+        },
+      },
+      data: { position: index + 1 },
+    });
+  });
```
oleksandr-danylchenko: This can be optimized with parallel updates:
Use `await Promise.all((complectationId, index) => prisma.comparisons_Complectation.update....)`

In that way inside the `Promise.all` you'll have an array of promises that will be resolved in parallel

oleksandr-danylchenko (dismissed):

R0ixy:
Yesterday another branch was merged into this one and now after force push all commits disappeared. It looks like something went really wrong...

PR #229, opened 2022-09-07
Fix/edit profile

AlexanderKhapchenko:


backend/src/controllers/update-user/update-user.controller.ts
```diff
@@ -92,22 +100,49 @@ const updateUserPhoto = async (
...
+    });
+    const writableStream = new stream.PassThrough();
+    readableStream.pipe(transformer).pipe(writableStream);
 
-    const photoUrl = await uploadFileToS3(req.files.photo);
+    const photoUrl = await uploadFileToS3(writableStream, s3Key);
+    await UsersHelper.deleteUserPhoto(req.body.tokenPayload.sub);
     await userService.updateUserPhoto(req.body.tokenPayload.sub, photoUrl);
+
     return res.status(httpStatus.OK).json({ photoUrl });
```
oleksandr-danylchenko: What does the service layer logic do in the controller? :eye::eye:

backend/src/controllers/update-user/update-user.controller.ts
```diff
@@ -92,22 +100,49 @@ const updateUserPhoto = async (
...
   }
 };
 
+const deleteUserPhoto = async (
+  req: TypedRequestBody<{ tokenPayload: TokenPayload }>,
+  res: Response<Partial<UpdateUserReq>>,
+  next: NextFunction,
+): Promise<Response | undefined> => {
+  try {
+    await UsersHelper.deleteUserPhoto(req.body.tokenPayload.sub);
```
oleksandr-danylchenko: variables must be named in `camelCase`

oleksandr-danylchenko (changes requested):

PR #233, opened 2022-09-08
connect-the-characteristics-menu-with-the-backend

stanislav-ostr:


oleksandr-danylchenko (approved):

oleksandr-danylchenko:
There're more code-style comments I would like to leave. Especially for using the indices to access array entries. But as it fixes the issue and we don't have that much time left, this PR looks good enough.

PR #235, opened 2022-09-08
setting-the-carousel-to-detailed-page-layout

stanislav-ostr:


frontend/src/components/common/carousel/styles.scss
```diff
@@ -1,21 +1,18 @@
...
-
-.swiper-slide img {
-  display: block;
-  width: 100%;
-  height: 100%;
-  object-fit: cover;
+.mySwiper,
+.mySwiper2 {
+  .swiper-slide img {
+    display: block;
```
oleksandr-danylchenko: C'mon! Why do you name classes like `mySmth`? Class should describe what it represents on the page

oleksandr-danylchenko (changes requested):

frontend/src/components/common/carousel/styles.scss
stanislav-ostr: Oops! I will fix this

frontend/src/components/common/carousel/carousel.tsx
```diff
@@ -10,13 +10,12 @@ import 'swiper/css';
 import 'swiper/css/free-mode';
 import 'swiper/css/navigation';
 import 'swiper/css/thumbs';
-import { images } from './mock-data';
 
 import type { Swiper as SwiperType } from 'swiper';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
oleksandr-danylchenko: We don't need this disabled comment anymore

oleksandr-danylchenko (dismissed):

PR #246, opened 2022-09-08
Add recent search cars section

brovarenko:
See - https://app.asana.com/0/1202525679663009/1202919442667727/f
Capture


backend/src/services/recent-search-cars/resent-search-cars.service.ts
```diff
@@ -0,0 +1,105 @@
...
+      photoUrls: car.model.photo_urls,
+      brand: {
+        name: car.model.brand.name,
+        logoUrl: car.model.brand.logo_url,
+      },
+      pricesRanges: car.model.prices_ranges,
+      description: car.model.description,
+    } as RecentSearchCarsResponse;
+    cars.push(data);
+  });
```
oleksandr-danylchenko: Please re-read what the `.map` does :sweat_smile:. **It returns a new array.** Having another array that we populate via push is totally redundant here.

It's also a good practice to not have any side effects inside the immutable methods like `.map`, `.filter` and others

oleksandr-danylchenko (changes requested):

backend/src/services/recent-search-cars/resent-search-cars.service.ts
brovarenko: I didn't notice, I took this piece of code from our project

oleksandr-danylchenko: oh-oh... So that piece of code needs to be re-written too

oleksandr-danylchenko (approved):

PR #248, opened 2022-09-08
create detailed page

stanislav-ostr:


frontend/src/components/details-page/collapse-block/collapse-block.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const { caption, children, open } = props;
+  const [collapseOpen, setCollapseOpen] = useState(open);
+
+  useEffect(() => {
+    setCollapseOpen(open);
+  }, [open]);
+
+  const toggleCollapse = (): void => {
+    setCollapseOpen(!collapseOpen);
+  };
```
oleksandr-danylchenko, line 23: How should the parent know if the CollapseBlock was closed? Because I clearly see that the local `collapseOpen` can out-of-sync with the `props.open`

oleksandr-danylchenko (changes requested):

frontend/src/components/details-page/collapse-block/collapse-block.tsx
stanislav-ostr: The 'open' prop is needed in two cases:
* setting the initial state
* forced opening of the collapse when user choosing a configuration
* in other cases, the user can independently regulate the collapse without informing the parent

oleksandr-danylchenko (approved):

PR #249, opened 2022-09-09
Feature/autotests

Elytr0n:
Added


oleksandr-danylchenko (approved):

oleksandr-danylchenko:
@RomanMahola 
Could you look at the code and merge it if it would be appropriate? Because I'm not a big expert in cypress)

PR #252, opened 2022-09-09
make sequential notifications

ashonuyak:
See - https://app.asana.com/0/1202525679663009/1202945039404865/f

image

### DIFF
--- a/frontend/src/contexts/wishlist-context.tsx
+++ b/frontend/src/contexts/wishlist-context.tsx
@@ -58,7 +58,7 @@ const WishlistContextProvider: React.FC<{ children: ReactNode }> = ({
     if (isLiked) {
       const existing = notifications.find(
         (n) =>
-          n.modelId === data.modelId ??
+          n.modelId === data.modelId &&
           n.complectationId === data.complectationId,
       );
       const nextNotifications = existing
--- a/frontend/src/contexts/wishlist-context.tsx
+++ b/frontend/src/contexts/wishlist-context.tsx
@@ -19,9 +19,19 @@ const WishlistContext = createContext<WishlistContextType>({
   handleLikeClick: () => undefined,
 });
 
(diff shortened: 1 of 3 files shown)


frontend/src/contexts/wishlist-context.tsx
```diff
@@ -44,33 +54,96 @@ const WishlistContextProvider: React.FC<{ children: ReactNode }> = ({
     const isLiked = likedCars?.includes(carId as string);
 
     isLiked ? handleDeleteWishlist(data) : handleCreateWishlist(data);
+
+    if (isLiked) {
+      const existing = notifications.find(
+        (n) =>
+          n.modelId === data.modelId ??
+          n.complectationId === data.complectationId,
+      );
```
oleksandr-danylchenko: Why do you use `??` here? `n.modelId === data.modelId` always returns **a defined** boolean

oleksandr-danylchenko (dismissed):

frontend/src/contexts/wishlist-context.tsx
ashonuyak: it seems to me that we sometimes like a model of the car and sometimes an exact complectation of the car, so thats why we have such interface for liked cars 
image

oleksandr-danylchenko: No, I mean that exactly the `??` is redundant here. Because ` n.modelId === data.modelId ?? n.complectationId === data.complectationId` is getting compiled to the `boolean ?? boolean`. But `??` is used as the `&&` operator but it doesn't return false if on the left side you'll have a falsy value (`''`, `0`, `NaN`). In our case `&&` would be more appropriate.

Also, if you target your TS compiler to ES5, the `??` will generate much more code because of the transpilation: 
image

oleksandr-danylchenko (approved):

PR #254, opened 2022-09-09
fix max password length

ashonuyak:
See - https://app.asana.com/0/1202525679663009/1202952266350607/f


shared/src/validation-schemas/forgot-password/forgot-password.validation-schema.ts
```diff
@@ -1,6 +1,9 @@
 import * as Yup from 'yup';
 
-import { UserValidationMessage } from '../../common/enums/enums';
+import {
+  UserValidationMessage,
+  UserValidationRule,
+} from '../../common/enums/enums';
```
oleksandr-danylchenko: Alias?

oleksandr-danylchenko (dismissed):

PR #256, opened 2022-09-09
Fixed notification position by added header offset

AlexanderKhapchenko:


AlexanderKhapchenko:
image
image

Or maybe it would be better if the notification message moved to the bottom?
image

oleksandr-danylchenko:
Let's show them at the bottom. It would be easier to manage the offset

oleksandr-danylchenko (commented):
Let's show them at the bottom. It would be easier to manage the offset

Thanks for the suggestion :raised_hands:

PR #258, opened 2022-09-09
Replaced header bell icon

iryb:
Bell icon was replaced with eye, like the icon we have on Personal page for viewed cars.

image

This link leads to Viewed cars section of Personal page.

I also added smooth scroll to section.

Task https://app.asana.com/0/1202525679663009/1202945486360273


frontend/src/assets/css/styles.scss
```diff
@@ -14,6 +14,10 @@
   box-sizing: border-box;
 }
 
+html {
+  scroll-behavior: smooth;
+}
```
oleksandr-danylchenko: Modifying `html` is considered a bad practice. Does this property have any other side-effects except making the "to-anchor" scroll smoother?

iryb: I got it

oleksandr-danylchenko (approved):

PR #259, opened 2022-09-09
fix bug 71 edit profile start/end years

setar92:
https://app.asana.com/0/1202525679663009/1202951564209238


frontend/src/components/edit-profile/edit-profile.tsx
```diff
@@ -224,7 +224,7 @@ export const EditProfile: React.FC<EditProfileProps> = ({ onClose }) => {
                     <MenuItem value="not_appliable">Rather not say</MenuItem>
                   </SelectFieldForm>
                   <SelectYearRange
-                    start={1960}
+                    start={new Date().getFullYear() - 95}
```
oleksandr-danylchenko: This should be in some variable. Firstly - to be more explicit. Secondly, on each rerender a new date object will be created

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #260, opened 2022-09-09
Add comparison page

R0ixy:
https://app.asana.com/0/1202525679663009/1202943623164029/f

image

### DIFF
--- a/frontend/src/components/collapse-component/collapse-element/styles.module.scss
+++ b/frontend/src/components/collapse-component/collapse-element/styles.module.scss
@@ -15,7 +15,7 @@
 
 .show {
   height: auto;
-  max-height: 1500px;
+  max-height: 2000px;
   transition-duration: 0.8s;
 }
 
--- a/frontend/src/components/comparison-page/comparison-page.tsx
+++ b/frontend/src/components/comparison-page/comparison-page.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+
+import { CompTopTableBar } from '@components/comp-top-table-bar/comp-top-table-bar';
+import { OptionsSubtable } from '@components/comparison-options-table/options-subtable';
+import { GeneralComparisonTable } from '@components/general-comparison-table/general-comparison-table';
+import { Header } from '@components/header/header';
+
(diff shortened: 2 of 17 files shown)


frontend/src/components/collapse-component/collapse-element/styles.module.scss
```diff
@@ -15,7 +15,7 @@
 
 .show {
   height: auto;
-  max-height: 1500px;
+  max-height: 2000px;
   transition-duration: 0.8s;
 }
```
oleksandr-danylchenko, line 20: Why `2000px`?

frontend/src/components/comparison-page/comparison-page.tsx
```diff
@@ -0,0 +1,40 @@
...
+            <CompTopTableBar />
+            <GeneralComparisonTable />
+            <OptionsSubtable title="design" />
+            <OptionsSubtable title="sound" />
+            <OptionsSubtable title="security" />
+            <OptionsSubtable title="multimedia" />
+            <OptionsSubtable title="comfort" />
+            <OptionsSubtable title="optics" />
+            <OptionsSubtable title="auxiliary" />
+            <OptionsSubtable title="upholstery" />
```
oleksandr-danylchenko: I believe that it can be done more elegantly using the array of options and `.map`

oleksandr-danylchenko (approved):

frontend/src/components/collapse-component/collapse-element/styles.module.scss
R0ixy: We have dynamic tables height and 1500px was not enough on some screen sizes

oleksandr-danylchenko: Okay. Why do we need to have `max-height` at all?

oleksandr-danylchenko (approved):

frontend/src/components/collapse-component/collapse-element/styles.module.scss
R0ixy: That's an interesting question, but collapse component don't work without it

PR #261, opened 2022-09-09
Fixed incorrect search results when selecting two brands with model

AlexanderKhapchenko:
see:
https://app.asana.com/0/1202525679663009/1202951564209228/f


backend/src/services/cars/cars-local-search.service.ts
```diff
@@ -2,6 +2,24 @@ import { CarsSearchParams, SearchResult } from '@autoline/shared';
...
+          modelIds.length && {
+            id: {
+              in: modelIds,
+            },
+          }),
+        brand: {
+          id: brandId,
+        },
+      })),
+    ],
```
oleksandr-danylchenko: You already create a new array with the `.map`. Why to you spread it again into an empty array?

frontend/src/helpers/object-to-query.ts
```diff
@@ -1,16 +1,52 @@
...
+  [p: string]: string | string[] | Record<string, string | string[]>[];
 }): string[][] => {
   const notEmpties = Object.fromEntries(
     Object.entries(filters).filter(
-      ([_, value]) => value.length >= 1 && value !== '' && value[0] !== '',
+      ([_, value]) =>
+        value.length >= 1 &&
+        value !== '' &&
+        Array.isArray(value) &&
+        value[0] !== '',
```
oleksandr-danylchenko: > `value.length >= 1 && value !== ''`

:eye::eye:

Also, you don't need to make a comparison with the `''`. You can just check `!value` or `!value[0]`

frontend/src/helpers/object-to-query.ts
```diff
@@ -1,16 +1,52 @@
...
 
   return Object.entries(notEmpties).flatMap(([key, value]) => {
     if (typeof value === 'string') {
       return [[key, value]];
     }
-    return value.map((item) => [key, item]);
+    if (
+      Array.isArray(value) &&
+      (value as []).every((item) => typeof item === 'object')
+    ) {
```
oleksandr-danylchenko: `as []`

:eye::eye:

frontend/src/helpers/object-to-query.ts
```diff
@@ -1,16 +1,52 @@
...
+      return objectsArrayToQuery(
+        key,
+        value as {
+          [p: string]: string | string[];
+        }[],
```
oleksandr-danylchenko: `as { [p: string]: string | string[]; }[]`

:eye::eye:

oleksandr-danylchenko (changes requested):

frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
```diff
@@ -112,8 +112,10 @@ const AdvancedAutoFilter: FC = () => {
       objectToQueryString({
         ...rangeFiltersToObject(rangeFilters),
         ...checkLists,
-        brandId: brandDetails.map((item) => item.brandId),
-        modelId: brandDetails.flatMap((item) => item.modelIds),
+        brandDetails: brandDetails.map(({ modelIds, brandId }) => ({
+          modelIds,
+          brandId,
+        })),
```
elador13: key `brandDetails`  doesn't exist on [specification](https://docs.google.com/document/d/16dU-biODvptcGAsv2I6EBGBKiQkJLGKvraUJYM4CkcU/edit) rules. I think you can check brand details in filters and: if in brand details only `brandId `- pass `brandId` in query, if also present `modelIds` - pass only `modelIds `without their `brandId`. I think it should work, won't break spec. rules and could be simple to handle on the back-end side. Because previously I passed in query `brandId ` and its `modelIds` together. So my suggestion is to pass only `modelIds`, without their `brandId` (if it's possible of course)

frontend/src/helpers/object-to-query.ts
AlexanderKhapchenko: image
Without it TS thinks that it's not an array
Although this is strange, because there is an Array.isArray(value) check

frontend/src/helpers/object-to-query.ts
AlexanderKhapchenko: Again, I don't understand why the Array.isArray(value) check doesn't work
Because TS doesn't understand what's really going on

elador13 (approved):
Looks good!

oleksandr-danylchenko (approved):

PR #262, opened 2022-09-09
Fixed active state of navigation element in site header

AlexanderKhapchenko:


frontend/src/components/header/header.tsx
```diff
@@ -25,7 +25,8 @@ import { PrivateElements } from './private-elements/private-elements';
 import styles from './styles.module.scss';
 
 export const Header = (): React.ReactElement => {
-  const [value, setValue] = useState(0);
+  const [value, setValue] = useState(-1);
```
oleksandr-danylchenko: What? What does the `value` mean? Why it's even a `number`?

frontend/src/components/header/header.tsx
```diff
@@ -47,6 +48,16 @@ export const Header = (): React.ReactElement => {
...
 
+  useEffect(() => {
+    if (location.pathname === AppRoute.SEARCH) {
+      setValue(0);
+    } else if (location.pathname === AppRoute.ABOUT_US) {
+      setValue(1);
+    } else {
+      setValue(-1);
+    }
+  });
```
oleksandr-danylchenko: This should be in the `switch` statement as we have comparisons of the static values. What does the numeric value even means? 

To display that smth active we need to store the active route string

oleksandr-danylchenko (changes requested):

iryb:
I think we need to refactor header menu to use react router dom's `NavLink`. It automatically adds some attributes to anchor tag when it matches the current URL.

oleksandr-danylchenko (approved):

PR #264, opened 2022-09-10
Make cars categories clickable

brovarenko:
See - https://app.asana.com/0/1202525679663009/1202943667066299/f

ezgif com-gif-maker (1)


frontend/src/components/cars-categories/cars-categories.tsx
```diff
@@ -13,7 +19,30 @@ import 'swiper/css/mousewheel';
...
+    dispatch(
+      setBrandDetailsValue({
+        id: detailId,
+        brandId: queryParams[0][1] || '',
+        modelIds: [],
+      }),
+    );
+    await search(queryParams);
+    navigate(API.SEARCH);
+  };
```
oleksandr-danylchenko: That's wrong. **The search page should be responsible for searching cars**. The categories slider should just provide search params for it

oleksandr-danylchenko (changes requested):

frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
```diff
@@ -63,6 +64,8 @@ import styles from './styles.module.scss';
 
 const AdvancedAutoFilter: FC = () => {
   const dispatch = useAppDispatch();
+  const [searchParams] = useSearchParams();
+  const brandId = (searchParams.get('brand') as string) ?? '';
```
oleksandr-danylchenko: Why do we need the `as string`? `.get` already returns the `string`: 
image

Also, why do we need `??` here instead of the simpler `||`?

frontend/src/components/advanced-auto-filter/advanced-auto-filter.tsx
```diff
@@ -128,6 +131,23 @@ const AdvancedAutoFilter: FC = () => {
...
+          modelIds: [],
+        }),
+      );
+      search([['brandId', brandId]]);
+      window.scrollTo({
+        top: 0,
+        behavior: 'smooth',
+      });
+    }
+  }, []);
```
oleksandr-danylchenko: Guard clause should be preferred here - `if (!brandId) return;`

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #266, opened 2022-09-10
fix/Bug_68 Show applied filters after search

elador13:
See - https://app.asana.com/0/1202525679663009/1202951564209223/f

### **Result:**
applied-bar-after-search


oleksandr-danylchenko (dismissed):
LGTM (Let's gamble, try merging)

PR #276, opened 2022-09-12
fix header tabs incorrect behaviour

ashonuyak:
See - https://app.asana.com/0/1202525679663009/1202960994890826/f

### DIFF
--- a/frontend/src/components/header/header.tsx
+++ b/frontend/src/components/header/header.tsx
@@ -26,7 +26,18 @@ import { PrivateElements } from './private-elements/private-elements';
 import styles from './styles.module.scss';
 
 export const Header = (): React.ReactElement => {
-  const [value, setValue] = useState(0);
+  let tabNumber;
+  switch (window.location.pathname) {
+    case AppRoute.SEARCH:
+      tabNumber = 0;
+      break;
+    case AppRoute.ABOUT:
+      tabNumber = 1;
+      break;
+    default:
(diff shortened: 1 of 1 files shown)


frontend/src/components/header/header.tsx
```diff
@@ -26,7 +26,18 @@ import { PrivateElements } from './private-elements/private-elements';
...
+  switch (window.location.pathname) {
+    case AppRoute.SEARCH:
+      tabNumber = 0;
+      break;
+    case AppRoute.ABOUT:
+      tabNumber = 1;
+      break;
+    default:
+      tabNumber = null;
+  }
```
oleksandr-danylchenko, line 39: Please, don't use the number to define the active tab element. Because it'll complicate changing the order of the tabs. Store unique identified of the tab - it's name

oleksandr-danylchenko (changes requested):

frontend/src/components/header/header.tsx
ashonuyak: @oleksandr-danylchenko but mui Tabs component receives only the number of the chosen tab. I can do it the following way, but we still need the number of tab 
image

oleksandr-danylchenko: @ashonuyak 
Then you need to use the `enum`. It's a perfect tool if you need to attach a human-readable name to some unknown numeric value in a predictable manner

PR #277, opened 2022-09-12
Fixed bug 79 guest users should be able to visit Search Page

R0ixy:
https://app.asana.com/0/1202525679663009/1202958674462101/f

### DIFF
--- a/frontend/src/store/queries/comparisons/index.ts
+++ b/frontend/src/store/queries/comparisons/index.ts
@@ -87,3 +87,6 @@ export const {
   useGetComparisonCarsQuery,
   useUpdatePositionsMutation,
 } = comparisonsApi;
+
+export const comparison =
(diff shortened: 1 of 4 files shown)


oleksandr-danylchenko (dismissed):

frontend/src/store/queries/comparisons/index.ts
```diff
@@ -87,3 +87,6 @@ export const {
   useGetComparisonCarsQuery,
   useUpdatePositionsMutation,
 } = comparisonsApi;
+
+export const comparison =
+  comparisonsApi.endpoints.addCarToComparison.matchRejected;
```
oleksandr-danylchenko: Please name it more explicitly. Because `comparison` is too generic and doesn't convey what it really means

oleksandr-danylchenko (approved):

PR #289, opened 2022-09-12
Fixed label in select field

AlexanderKhapchenko:


oleksandr-danylchenko:
@elador13 
This is your realm and would believe your approval when it'll be granted

PR #290, opened 2022-09-12
Live updates for comparison page

R0ixy:
https://app.asana.com/0/1202525679663009/1202961645454088/f


frontend/src/components/comp-top-table-bar/comp-top-table-bar.tsx
```diff
@@ -34,6 +34,13 @@ export const CompTopTableBar = (): React.ReactElement => {
     handleClearTable();
   };
 
+  const broadcast = new BroadcastChannel('compare');
+  useEffect(() => {
+    broadcast.onmessage = (): void => {
+      refetch();
+    };
+  }, [broadcast, refetch]);
```
oleksandr-danylchenko: This `BroadcastChannel` will be created on each render. Please move it either to the `useEffect` or `useMemo`. The first option is preferred as you don't use it outside of it.

Also, on component unmount we should close `BroadcastChannel`

oleksandr-danylchenko (changes requested):

frontend/src/components/comp-top-table-bar/comp-top-table-bar.tsx
R0ixy: It's very strange but it stops working at all if I'm adding this to useEffect    
```
return (): void => {
   broadcast.close();
};
```

oleksandr-danylchenko (approved):

PR #291, opened 2022-09-12
Hide empty options tables in only difference mode

R0ixy:
https://app.asana.com/0/1202525679663009/1202961645454074/f


frontend/src/components/comparison-page/comparison-page.tsx
```diff
@@ -8,10 +8,38 @@ import { CompTopTableBar } from '@components/comp-top-table-bar/comp-top-table-b
...
+    }
+
+    return useMemo(
+      () =>
+        complectationOptions.map((option) => (
+          <OptionsSubtable key={option} title={option} />
+        )),
+      [data],
+    );
+  };
```
oleksandr-danylchenko: It would be better to extract these 2 `useMemo` into 2 sub-components. It would be a much more idiomatic React solution

R0ixy: Do we need to wrap these sub-components in `React.memo()` or...
Why is it better? Won't they re-render each time user toggles table view?

PR #292, opened 2022-09-12
Fixed Bug_90 Add scroll to top after redirect to Detailed view

iryb:
Task https://app.asana.com/0/1202525679663009/1202961645454118

## Before
detailed-view-scroll-before (1)

## After
detailed-view-scroll-after (1)

### DIFF
--- a/frontend/src/components/car-list-item/car-list-item.tsx
+++ b/frontend/src/components/car-list-item/car-list-item.tsx
@@ -113,6 +113,7 @@ const CarListItem: React.FC<CarListItemProps> = (props) => {
 
   const handleCarModelClick = (): void => {
     navigate({ pathname: AppRoute.DETAILS, search: `?model=${model_id}` });
+    scrollTo(0, 0);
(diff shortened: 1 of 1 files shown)


oleksandr-danylchenko (changes requested):
For better incapsulation, it would be better to put the scroll restoration in the component that requires it - https://v5.reactrouter.com/web/guides/scroll-restoration

image

Also, I would put such `ScrollToTop` directly in the routing definition because then for other devs it would be crystal clear what pages scroll to the top and what not

iryb:
@oleksandr-danylchenko We have two scroll scenarios for Detailed view page. 

If user clicks on car title (Search page), Detailed page should be opened with scroll position on top. 
image

If he clicks on complectation row, Detailed page opens with scroll position on complectation details
image

Adding scroll to top as component inside Detailed page will force it to always be scrolled to top
image

oleksandr-danylchenko (approved):

PR #303, opened 2022-09-13
Fixed bug 95

AlexanderKhapchenko:
see:
https://app.asana.com/0/1202525679663009/1202970261868378/f

### DIFF
--- a/backend/src/services/cars/cars-local-search.service.ts
+++ b/backend/src/services/cars/cars-local-search.service.ts
@@ -47,7 +47,7 @@ const carsSearch = async (data: CarsSearchParams): Promise<SearchResult[]> => {
         {
           brand: {
             id: {
-              in: data.brandId,
+              in: data.brandId ? data.brandId : data.modelId ? [] : undefined,
             },
           },
         },


backend/src/services/cars/cars-local-search.service.ts
```diff
@@ -47,7 +47,7 @@ const carsSearch = async (data: CarsSearchParams): Promise<SearchResult[]> => {
         {
           brand: {
             id: {
-              in: data.brandId,
+              in: data.brandId ? data.brandId : data.modelId ? [] : undefined,
```
oleksandr-danylchenko, line 50: I don't understand ;(
"Don't include cars by brand if the model is present?"

elador13: Yes, we already pass certain model of brand, and if modelId present, we don't need its brandId.
It's a little unexpected code, but I tested and it works correctly.

oleksandr-danylchenko: Extremely weird, but lets just make it work before the demo :cry:

PR #307, opened 2022-09-13
fixed createdAt types

elador13:


shared/src/common/types/preferences/wishlist-response-dto.type.ts
```diff
@@ -14,6 +14,7 @@ interface WishlistsResponseDto {
 type WishlistInput = {
   modelId?: string;
   complectationId?: string;
+  createdAt: string;
```
iryb: I think `WishlistInput` shouldn't have the required `createdAt` property. It should be optional.

elador13: Already done

iryb (approved):

PR #308, opened 2022-09-13
Fix empty row in comparison table

brovarenko:
See - https://app.asana.com/0/1202525679663009/1202961645454070/f


frontend/src/helpers/comparsion/find-empty-option.ts
```diff
@@ -0,0 +1,17 @@
+import { ComparisonGeneralInform } from '@autoline/shared';
+
+const findEmptyOptions = (
+  generalInfo: ComparisonGeneralInform[] | undefined,
+): string[] | undefined => {
+  const emptyOptions = generalInfo?.map((e) =>
+    Object.entries(e.options)
+      .filter((option) => option[1].length == 0)
+      .map((option) => option[0]),
+  );
```
oleksandr-danylchenko: It's a great practice to destruct the `Object.entries` result instead of using the indices because they are unclear.
`.filter(([smthA, smthB]) => {})`

frontend/src/components/general-comparison-table/general-comparison-table.tsx
```diff
@@ -102,15 +103,18 @@ const GeneralComparisonTable: React.FC = () => {
...
-            <div
-              className={styles.tableCell}
-              data-optiontitle={option}
-              key={uuid4()}
-            >
-              {option}
-            </div>
-          ))}
+          {[...options].map((option) => {
+            if (findEmptyOptions(generalInfo)?.includes(option)) return;
```
oleksandr-danylchenko: Move `findEmptyOptions` outside the loop. Because now it redundantly reruns on each loop iteration

oleksandr-danylchenko (changes requested):

oleksandr-danylchenko (approved):

PR #309, opened 2022-09-13
Add <only difference> toggle

R0ixy:
https://app.asana.com/0/1202525679663009/1202961645454102/f


frontend/src/components/general-comparison-table/general-comparison-table.tsx
```diff
@@ -80,31 +117,46 @@ const GeneralComparisonTable: React.FC = () => {
...
-            Engine
-          </div>
-          <div className={styles.tableCell} data-optiontitle="wheeldrive">
-            Wheel Drive
-          </div>
+          {(toggle && compareFields.bodyType) || (
+            <div className={styles.tableCell} data-optiontitle="bodytype">
+              Type
+            </div>
+          )}
```
oleksandr-danylchenko: > If toggle is `true` -> show `compareFields.bodyType`, and if not -> show `div`

Do I understand correctly what it does?

Because if it's correct, I believe a simple `?:` operator would be much easier to comprehend

frontend/src/components/general-comparison-table/general-comparison-table.tsx
```diff
@@ -125,46 +179,70 @@ const GeneralComparisonTable: React.FC = () => {
...
+                    </div>
+                  )}
                   {Object.keys(info.options).map(
                     (type: string) =>
-                      !findEmptyOptions(generalInfo)?.includes(type) && (
+                      !findEmptyOptions(
+                        generalInfo,
+                        toggle,
+                        carOmitOptions,
+                      )?.includes(type) && (
```
oleksandr-danylchenko: We shouldn't run the `findEmptyOptions` on each iteration

oleksandr-danylchenko (changes requested):

frontend/src/components/general-comparison-table/general-comparison-table.tsx
R0ixy: Nope, I use brackets `()` and `&&` here so when both `toggle` and `compareFields.bodyType` are `true` -> hide `div`. In all other situations show `div`. 

`toggle` = `true` means we're in `only difference` mode now. 
`compareFields.bodyType` = `true` means bodyType value for all cars is the same. (Maybe I need to name it more explicit)

oleksandr-danylchenko: > toggle = true means we're in only difference mode now.

Then the `toggle` is extremely misleading name

oleksandr-danylchenko: > hide div. In all other situations show div.

`||` is an extremely unintuitive way to do that! Please rewrite it with the `&&`: `{(!isOnlyDiff || !compareFields.bodyType) && <div>...</div>}`

oleksandr-danylchenko (approved):
I want to believe that it works :pray:

PR #311, opened 2022-09-13
Simplify Equal height logic

iryb:
`react-equal-height` was added to simplify adjusting table rows height.

This should fix bug https://app.asana.com/0/1202525679663009/1202961645454055 that appears from time to time on staging.

Bug: 
image

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -32,6 +32,7 @@
     "clsx": "1.2.0",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "react-equal-height": "^1.2.2",
     "react-hook-form": "^7.34.2",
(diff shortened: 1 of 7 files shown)


frontend/package.json
```diff
@@ -32,6 +32,7 @@
     "clsx": "1.2.0",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "react-equal-height": "^1.2.2",
```
oleksandr-danylchenko, line 35: I like it! I would even recommend it to a colleague who's struggling w/ the same issue. Thanks :raised_hands:

oleksandr-danylchenko (approved):

PR #328, opened 2022-09-14
saved adverts for all complectations which user chosed

setar92:
Now we save adverts and page number for all compectations which user chosed. When user choses complectation second time - we use adverts from redux store.  
Bug https://app.asana.com/0/1202735397457083/1202978597528825
https://gyazo.com/88efbea349f8adf286c3e397b1e5162c


oleksandr-danylchenko (dismissed):

oleksandr-danylchenko:
@setar92 
Please fix the typescript error and then this PR will be merged

PR #344, opened 2022-09-16
fix/Bug_108 + Rewrite all notifications on react toastify

elador13:
See - https://app.asana.com/0/1202525679663009/1202978745534216/f

https://user-images.githubusercontent.com/49813216/190624206-2c4e861b-e70c-4f5d-a66d-a7dcdf80e31f.mp4

https://user-images.githubusercontent.com/49813216/190624574-48025533-38a5-4ac6-9665-7d03f14b292e.mp4


frontend/src/contexts/wishlist-context.tsx
```diff
@@ -12,123 +13,60 @@ import {
...
-      setNotifications([]);
-    } else {
-      const ids = Array.isArray(id) ? id : [id];
-      const nextNotifications = notifications.filter(
-        ({ modelId, complectationId }) =>
-          !ids?.includes(modelId || complectationId || uuid4()),
-      );
-      setNotifications(nextNotifications);
-    }
+    toast.info(
```
R0ixy: In case a user is unauth, he will be redirected to sign-in but the notification still appears. 

I think we need to put this call inside of the `createWishlist` promise using `unwrap().then()`

frontend/src/contexts/compare-context.tsx
```diff
@@ -10,105 +12,64 @@ import {
...
-              complectationId,
-              carName: name,
-              timestamp: new Date().getTime(),
-            });
-        setNotifications(nextNotifications);
-        broadcast.postMessage('compare');
-      });
+    addCarToComparison({ complectationId });
+
+    toast.info(<Notification children={`You added ${name} to comparison`} />, {
```
R0ixy: Same as I pointed out for the wishlist

R0ixy (changes requested):
By the way, if we add undo functionality for comparisons I think we need to add it to the delete from comparison button on the comparisons page as well

iryb:
@elador13 Please adjust styles of notification to follow mockup more accurately.

Original design:
image


1. Add more padding on top and bottom so text can have a breath.

image

2. Make the icon aligned by text.

image

3. Change the color of car title (it's underlined on screenshot) to blue. Other text should be black. Also, car title should have `text-transform: uppercase` property.

image

Thanks!

R0ixy (approved):
LGTM!

PR #350, opened 2022-09-16
added prices fileds to comparing table

setar92:
https://gyazo.com/203b344fd7f36a18c9f1246fc18a82dd


frontend/src/components/general-comparison-table/general-comparison-table.tsx
```diff
@@ -77,16 +85,34 @@ const GeneralComparisonTable: React.FC<{ isOnlyDiff: boolean }> = ({
...
-  if (isLoading) return <Spinner />;
+  }, [refetch, refetchPrices]);
+  if (isLoading && isLoadingPrices) return <Spinner />;
 
   return (
     <CollapseElement label="General information" isOpen={true}>
       <EqualHeight>
         <div className={clsx(styles.table, 'table')}>
           <div className={clsx(styles.tableTitles, styles.tableColumn)}>
+            {(!isOnlyDiff || !isIdentical.bodyType) && (
```
R0ixy: Why do you use `!isIdentical.bodyType` here? It means when `bodyType` is the same for all cars this row will not be displayed. I'm pretty sure that's not what we need to achieve here

setar92: You are right, it is mistake. Now it's corrected, we will show prices always.

iryb (approved):

PR #356, opened 2022-09-19
fields-layout-fix

stanislav-ostr:
Снимок экрана 2022-09-19 122526


elador13 (commented):
Looks good, but arrows still not inlined. 
fields

Also, model field is unstable when we select/unselect brand:
model-field


@iryb what do you think?

iryb:
Yes, we have some troubles with placeholders alignment.

iryb (approved):

PR #365, opened 2022-09-20
Change filtering by price

R0ixy:


iryb:
@o-rudych what do you think about this?

o-rudych (approved):
Looks like it works.

PR #378, opened 2022-09-21
Added border on hover in edit-profile

elador13:
See - https://app.asana.com/0/1202525679663009/1203012299917843/f

image


iryb:
@elador13 actually, the border is not missing here. See Edit profile modal window:
image

Here is design reference. If button doesn't have border initially, text color becomes darker on hover. 
image

My suggestion here is to make white space between two buttons unclickable. User should see cursor pointer only when he hovers the "Cancel" text.

iryb (approved):

PR #379, opened 2022-09-21
Display No options text if nothing in list

iryb:
Task https://app.asana.com/0/1202525679663009/1203004555114342

image

### DIFF
--- a/frontend/src/components/comparison-popup/comparison-popup.tsx
+++ b/frontend/src/components/comparison-popup/comparison-popup.tsx
@@ -252,6 +252,9 @@ const ComparisonPopup: FC<ComparisonPopupProps> = ({
                       defaultValue="not_appliable"
                       disabled={!watch('model') || !watch('brand')}
                     >
+                      {availableComplectation.length === 0 && (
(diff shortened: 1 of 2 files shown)


frontend/src/components/comparison-popup/comparison-popup.tsx
```diff
@@ -252,6 +252,9 @@ const ComparisonPopup: FC<ComparisonPopupProps> = ({
                       defaultValue="not_appliable"
                       disabled={!watch('model') || !watch('brand')}
                     >
+                      {availableComplectation.length === 0 && (
+                        <li className={styles.noOptions}>No options</li>
+                      )}
```
elador13: I think ternary condition would be little more understandable here: 

`{availableComplectation.length ? (
                        availableComplectation.map((item) => (
                          <MenuItem key={item.id} value={item.id}>
                            {item.name}
                          </MenuItem>
                        ))
                      ) : (
                        <li className={styles.noOptions}>No options</li>
                      )}`

However your solution also not bad.

elador13 (dismissed):

elador13 (approved):

PR #381, opened 2022-09-21
hided empty general info table  (page comparison)

setar92:
https://app.asana.com/0/1202525679663009/1203020763697218
https://gyazo.com/a98344791435df8882fa70a1193edad0


iryb:
@setar92 I like how it looks! Can we also hide "All parameters/Only differences" toggle if there's no car selected? Because this toggle becomes non functional then.

setar92:
@iryb https://gyazo.com/5ad848f381f5ac49f11ccab1ce82849c

iryb (approved):

